PR #134, opened 2012-11-01
unignore test/

sakari:
This is a merge request for 1.0.1. I seem to be unable to make make github see that...

When installing with npat=true the installation fails due to missing
files in test/. Fix is to unignore the test/ directory.

Bumped version in the hope of a minor release.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -4,5 +4,4 @@
 /assets/
 /coverage/
 /demo/
-/test/
 /tools/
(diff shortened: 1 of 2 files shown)


gr2m:
note that a downside of this is that the released package on npm is 1.1MB in size, 853,6kb is the test folder. I don’t know what `npat=true` is, is it still in use?

PR #234, opened 2014-03-27
Support async functions and the await keyword, pending standardization

benjamn:
Although the specification of `async` functions is not universally agreed upon yet, I thought I would implement what seems to be [the leading proposal](http://wiki.ecmascript.org/doku.php?id=strawman:async_functions), in an attempt to see if there are any difficulties involved in parsing the new syntax.

In other words, feedback is the primary goal of this pull request, and I do not expect it to be merged any time soon. I will keep rebasing it as long as anyone cares.

This pull request adds support for `async` function declarations, named and anonymous function expressions, and arrow function expressions.

The presence of the `async` keyword before a function causes the AST node for that function to contain the property `"async": true`. For backwards compatibility, the absence of an `async` keyword produces an AST node with no `"async"` property (rather than `"async": false`, which would also make sense but would require many more changes to existing tests).

I have also included support for both the standard `await` keyword and the undecided-upon `await*` keyword. For now, the `*` simply causes the `AwaitExpression` to contain the property `"all": true`, because the leading proposal seems to be that `await* iterable` should be sugar for `await Promise.all(iterable)`. If a different consensus emerges, it will be easy to change the property name.

The `await` keyword is illegal outside of an `async` function, similar to `yield`, but (unlike `yield`) it is legal for an `async` function to contain no `await` expressions.

### DIFF
--- a/esprima.js
+++ b/esprima.js
@@ -213,7 +213,6 @@ parseYieldExpression: true
         IllegalBreak: 'Illegal break statement',
         IllegalDuplicateClassProperty: 'Illegal duplicate property in class definition',
         IllegalReturn: 'Illegal return statement',
-        IllegalYield: 'Illegal yield expression',
         IllegalSpread: 'Illegal spread element',
         StrictModeWith:  'Strict mode code may not include a with statement',
         StrictCatchVariable:  'Catch variable may not be eval or arguments in strict mode',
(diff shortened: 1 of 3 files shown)


mikesherov:
@benjamn can you rebase this and sign the cla (if you haven't already)? http://contribute.jquery.org/CLA/  Thanks again for contributing as much as you have!

ariya:
This seems to be ES7 feature. @jeffmo @benjamn What's the plan with tracking this as we converge master and harmony branch? Should this be landed as is in the harmony branch?

mikesherov:
@ariya, might be worthwhile to fold this topic into our discussion of language level flags.

jeffmo:
I would be in favor of landing this in the harmony branch for now. Its still a pretty early-stage proposal, but thats all the more reason for people to begin experimenting with it (and you need a parser to do that).

I feel like longer term, once master is more a parity with harmony, it would be nice to get rid of the harmony branch in favor of option flags for experimental features; Even if it's a coarse-grained option -- if there's still the concern that fine-grained options are going to be a maintenance burden.

jeffmo:
cc @bterlson who has taken over as champion for the sync/await proposal

ariya:
@jeffmo I am fine with landing this in the harmony branch. Feel free to make adjustment (rebase etc) as needed and then merge it in.

ariya:
We need to track the syntax, early errors, etc. I filed #1079 as the placeholder for that kind of info.

ariya:
Closing this as wontfix (see #1172 for details).

PR #277, opened 2014-08-18
Correct non-standard jslint directive.

xiemaisi:
This fixes https://code.google.com/p/esprima/issues/detail?id=580.


ariya:
JSLint won't be used anymore (https://code.google.com/p/esprima/issues/detail?id=623). Thank you!

PR #285, opened 2014-09-19
Harmony: Fat arrow and destructuring in params

Naddiseo:
Currently esprima throws an exception in cases where the arguments to a fat arrow function have destructuring:

```javascript
fn( ({ field_name: field_value}) => field_value); //'Unexpected token =>',
```

This PR will likely fix [issue 554][0] as well.

Note: this patch is from a downstream project that I saw doesn't have this problem: https://github.com/termi/es6-transpiler/blob/master/lib/esprima_harmony.js#L3390-L3394

[0]: https://code.google.com/p/esprima/issues/detail?id=554


ariya:
Superseded by PR #299.

PR #286, opened 2014-09-26
Remove/fix duplicate class method check

fkling:
[Rev 26 of the ES6 spec](http://esdiscuss.org/topic/new-rev26-es6-draft-now-available), removes duplicate class method checks, except for "constructor".

This diff removes the dupe check from `parseMethodDefinition` and
implements it in `parseClassBody` for "constructor" only. I chose to still
keep track of static and prototype method names, since the spec
describes such a list. I moved the check to `parseClassBody` with the
intention to be able to reuse `parseMethodDefinition` for object literals
in the future.

I changed the existing tests which previously expected errors to expect
ASTs now. I also added new tests for duplicate `constructor` properties.

`npm test` runs fine and there is no benchmark or coverage regression.

Issues covered by this PR:
https://code.google.com/p/esprima/issues/detail?id=594

### DIFF
--- a/esprima.js
+++ b/esprima.js
@@ -216,6 +216,7 @@ parseYieldExpression: true
         IllegalContinue: 'Illegal continue statement',
         IllegalBreak: 'Illegal break statement',
         IllegalDuplicateClassProperty: 'Illegal duplicate property in class definition',
+        IllegalClassConstructorProperty: 'Illegal constructor property in class definition',
         IllegalReturn: 'Illegal return statement',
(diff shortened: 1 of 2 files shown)


ariya:
@fkling Would you mind rebasing this against the latest harmony branch? Thanks!

fkling:
Done! Do you think I added too many tests?

ariya:
Looks good, thank you! More tests are fine; the more, the merrier :-)

PR #292, opened 2014-11-05
Change where comments are attached back to 1.2 branch location.

briandipalma:
https://code.google.com/p/esprima/issues/detail?id=607

This logic is the same as the 1.2 branch stack eating logic, it results in
comment nodes being placed where tools like ESLint expect them to exist.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -3754,6 +3754,67 @@ var testFixture = {
             }
         },
 
+        'var o = {\n/** Desc */\nfoo: function(){}\n};': {
+            "type": "Program",
+            "body": [{
+                "type": "VariableDeclaration",
+                "declarations": [{
+                    "type": "VariableDeclarator",
+                    "id": {
+                        "type": "Identifier",
+                        "name": "o",
+                        "range": [4, 5]
+                    },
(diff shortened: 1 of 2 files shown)


test/test.js
```diff
@@ -3754,6 +3754,67 @@ var testFixture = {
             }
         },
 
+        'var o = {\n/** Desc */\nfoo: function(){}\n};': {
```
ikarienator: It's unfortunate that this file is not nearly a JSON file. We are currently using a JavaScript style here and it normally uses single quotes, without quoting the keys, and curly braces are on the next line. You can easily pick up the style from the tests around you.

briandipalma: No worries, you didn't. It seems that this issue is closed https://code.google.com/p/esprima/issues/detail?id=607 with AsExpected. Will I close this PR?

mikesherov:
Don't close this yet.

mikesherov:
@briandipalma thanks again for contributing! We just finished having the discussion over in #1024, and decided we're not going to go back, so I'm going to close this PR. Thanks again.

PR #297, opened 2014-12-09
Fix division by "this"

ef4:
`this / 100` is legal Javascript, and a variation of it actually appears
in moment.js.

Without this fix, esprima treats it as an un-terminated regular
expression.

https://code.google.com/p/esprima/issues/detail?id=616

### DIFF
--- a/esprima.js
+++ b/esprima.js
@@ -1344,7 +1344,7 @@ parseStatement: true, parseSourceElement: true */
             }
             return collectRegex();
         }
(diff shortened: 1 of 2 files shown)


ariya:
@ef4 Got my email?

ef4:
@ariya Oh yes, reply sent just now.

ariya:
Landed as https://github.com/ariya/esprima/commit/3920416089. Thank you!

PR #299, opened 2014-12-19
Fix destructuring and default arguments in arrow functions

fkling:
There are a couple of issues with using parameter destructuring and default arguments in arrow functions. For example these don't parse:

```
foo(([a,b]) => 42);
foo((a=1) => 42);
```

There is already an open pull request ( #285 ), but it incorrectly allows for default arguments and patterns to be declared without parenthesis, i.e. those would be allowed

```
foo([a,b] => 42);
foo(a=1 => 42);
```

This fixes [554](https://code.google.com/p/esprima/issues/detail?id=554) and [581](https://code.google.com/p/esprima/issues/detail?id=554).

### DIFF
--- a/esprima.js
+++ b/esprima.js
@@ -3299,7 +3299,8 @@ parseYieldExpression: true
     }
 
     function parseAssignmentExpression() {
-        var marker, expr, token, params, oldParenthesizedCount;
+        var marker, expr, token, params, oldParenthesizedCount,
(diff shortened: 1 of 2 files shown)


fkling:
@ariya: I just replied.

ariya:
Landed as https://github.com/ariya/esprima/commit/0e4014d2. Thank you!

PR #300, opened 2015-01-07
Fixed lint directive, Removed missing globals from lint directives, removing unused variable.

PlNG:
JSHint also flags parseLeftHandSideExpression and parseFunctionExpression as unused.

I could probably further reduce the need for globals with a little code reorganization.


ariya:
We are deprecating the use of JSLint, see https://code.google.com/p/esprima/issues/detail?id=623.

PR #301, opened 2015-01-09
[Harmony] Fix eslint errors and update .eslintrc

danez:
Updated .esltinrc file from master to be compatible to newest eslint version
Also fixed eslint consistent-return-type rule

Otherwise the test always failed


danez:
Ok seems I only locally have an issue, that npm test is using a newer eslint version as installed.

PR #302, opened 2015-01-09
[Harmony] Fix token errors/duplicates with Regexp

danez:
This should fix the issue that regular expressions are messing around with tokens
see https://code.google.com/p/esprima/issues/detail?id=440
jscs-dev/node-jscs#685

I'm not sure what I did exactly, but as far as I understand peek() is also adding a token for the next token and lookahead2() is only gathering linenumbers etc. if in "tokenmode"

I'm not aware about any side effects. The tests ran through, I only had to change two tests, as they were expecting the wrong order.

This is the example I tested with:

var abc = /([A-Z])/g;

Before:

```json
[
    {
        "type": "Keyword",
        "value": "var"
    },
    {
        "type": "Identifier",
        "value": "abc"
    },
    {
        "type": "Punctuator",
        "value": "="
    },
    {
        "type": "Punctuator",
        "value": ";"
    },
    {
        "type": "RegularExpression",
        "value": "/([A-Z])/g",
        "regex": {
            "pattern": "([A-Z])",
            "flags": "g"
        }
    },
    {
        "type": "Punctuator",
        "value": ";"
    }
]
```

After:

```json
[
    {
        "type": "Keyword",
        "value": "var"
    },
    {
        "type": "Identifier",
        "value": "abc"
    },
    {
        "type": "Punctuator",
        "value": "="
    },
    {
        "type": "RegularExpression",
        "value": "/([A-Z])/g",
        "regex": {
            "pattern": "([A-Z])",
            "flags": "g"
        }
    },
    {
        "type": "Punctuator",
        "value": ";"
    }
]
```

### DIFF
--- a/esprima.js
+++ b/esprima.js
@@ -1274,7 +1274,7 @@ parseYieldExpression: true
     }
 
     function scanRegExp() {
-        var str, ch, start, pattern, flags, value, classMarker = false, restore, terminated = false, tmp;
+        var str, ch, start, pattern, flags, value, classMarker = false, restore, terminated = false, tmp, token;
 
         lookahead = null;
         skipComment();
@@ -1387,10 +1387,8 @@ parseYieldExpression: true
             value = null;
         }
 
-        lookahead2();
-
         if (extra.tokenize) {
-            return {
+            token = {
                 type: Token.RegularExpression,
                 value: value,
                 regex: {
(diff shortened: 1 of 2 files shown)


danez:
Seems the range and location are still messed up, gona have a look at that tomorrow.

danez:
Ok I finally found a way to solve all the problems in all cases (parse+tokens, parse w/o tokens, tokenize)

ariya:
I appreciate the effort to investigate this. The real solution is however involving synchronizing the handling of token and regex between the harmony branch and the master branch. I'll land the fix pretty soon.

danez:
Thanks.

ariya:
For the reference, the forwardported fix covers the commit https://github.com/ariya/esprima/commit/fc79b9d827 and several others before that. See if that fits into your independent path of investigation as well. Thank you!

danez:
Yes thanks it works now,
https://code.google.com/p/esprima/issues/detail?id=440 seems to be fixed for me now.

PR #306, opened 2015-01-26
Introduce optional SAX for esprima.tokenize

aslushnikov:
The patch attempts to introduce saxCallback option to the
esprima.tokenize method. The option accepts a function
which serves as a callback for every token. The SAX
interface doesn't require an array to store all the tokens
and saves >50% of time when tokenizing ~2MB of javascript.

Performance comparison could be viewed here:
https://gist.github.com/aslushnikov/184b8be792373fee59c9.

Summary: the larger input javascript, the more performance gains
attained by the SAX api.

https://code.google.com/p/esprima/issues/detail?id=626

### DIFF
--- a/esprima.js
+++ b/esprima.js
@@ -1321,9 +1321,29 @@
                 range: [pos, index],
                 loc: loc
             });
+            if (extra.saxCallback) {
+                resolveOpenBracketTokens();
+                if (extra.tokens.length > extra.maxTokensBufferLength) {
+                    extra.tokens = extra.tokens.splice(-2);
+                }
+                extra.saxCallback.call(null, filterTokenLocation(extra.tokens[extra.tokens.length - 1]));
+            }
         }
 
         return regex;
+    }
+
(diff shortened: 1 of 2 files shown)


aslushnikov:
Any updates on this?

ariya:
Introducing a new public API must be carried out with a lot of consideration. Once it is out, it is almost impossible to take it back. Among others, this means some thoughtfulness w.r.t to the naming. For example, I think SAX is a bit unfortunate since it stands for Simple API for XML. Also, the use of Boolean modifier is likely to cause a confusion, see my blog post on [Boolean traps](http://ariya.ofilabs.com/2012/06/detecting-boolean-traps-with-esprima.html).

However, the biggest question we need to solve it whether the tokenizer speed is something that can be fixed or not. I think it's better to hold this off until there is more profiling result from the tokenizer itself.

ariya:
Let's hold this off for now and continue the discussion on this issue: #1019.

PR #315, opened 2015-02-06
Add Maintainer's Guide stub with feature proposal section

mikesherov:


mikesherov:
@jeffmo @ariya just trying to codify some of what we do as maintainers of the project.

ariya:
Should we use the wiki for this? It will reduce the commit clutter.

mikesherov:
We can go either way. I don't have a strong preference but just wanted to lay out a few arguments for keeping it in the actual repo:

1. Some people may actually read it that way. Very few people read the wikis, but again, not opposed to it.
2. Having a pull request means we get to have discussions and peer review of our policies.
3. Having md files like this means eventually we can build the website from the md files, like we do in JSCS. *This means that the website and the MD files in the repo can be kept in sync easily*. We should only have to point people to the website or the md files to get the same info. 
4. the commit traffic should be relatively low.

That said, if you want Wiki, I am fine with that too.

dmethvin:
I enabled the Wiki if you want to use it.

mikesherov:
OK, let's actually start at Wiki. We can always export MD files later if we want.

ariya:
@mikesherov I'd like the idea of generated documentation. I think we could continue to use the wiki as the sandbox/brainstorming place and eternalize it later as the .md file.

Issue #1015, opened 2015-02-09
Travis CI should run the tests with Node.js 0.12 and io.js


ariya:
Now that Node.js 0.12 has been [released](http://blog.nodejs.org/2015/02/06/node-v0-12-0-stable/), that is what should be used instead of 0.11.

In addition, io.js should be tested as well.

michaelficarra:
There are still a lot of people running `0.11`. You should keep that one around for a while until more people adopt `0.12`.

PR #1016, opened 2015-02-09
Run Travis CI tests with Node.js 0.12 (stable) and io.js.

ariya:
See #1015.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,9 @@
+sudo: false
 language: node_js
 install: "npm update -g npm; npm install"
 node_js:
+  - iojs
+  - "0.12"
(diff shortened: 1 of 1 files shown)


michaelficarra:
LGTM. Remove `0.8`?

arthurvr:
I would remove 0.8 too, but no big deal for now. I'm wondering if [container based infrastructure using `sudo: false`](http://docs.travis-ci.com/user/workers/container-based-infrastructure/) is an option for us, makes the build much faster.

mathiasbynens:
:+1: to adding `sudo: false` for speedier builds.

ikarienator:
:+1:

Issue #1031, opened 2015-02-13
Transition to TryStatement single handler


ariya:
To facilitate the transition of `TryStatement` from the old `handlers` to the new `handler` (see #1030), Esprima should produce both properties. Once all the major downstream tools have been ported to use the `handler` only, `handlers` can be safely removed in the next major version.

In the current master branch:

```js
> require('./esprima').parse('try{}catch(e){}')
{ type: 'Program',
  body: 
   [ { type: 'TryStatement',
       block: [Object],
       guardedHandlers: [],
       handlers: [Object],
       finalizer: null } ] }
```

Once this transition step is implemented, it will look like:

```js
> require('./esprima').parse('try{}catch(e){}')
{ type: 'Program',
  body: 
   [ { type: 'TryStatement',
       block: [Object],
       guardedHandlers: [],
       handlers: [Object],
       handler: [Object],
       finalizer: null } ] }
```

PR #1034, opened 2015-02-14
Trystatement handler

ariya:
See #1031 for details. `TryStatement` also contains `handler` to help the migration until 3.x is released.


esprima.js
```diff
@@ -1948,11 +1948,12 @@
...
         },
 
-        finishTryStatement: function (block, guardedHandlers, handlers, finalizer) {
+        finishTryStatement: function (block, handler, finalizer) {
             this.type = Syntax.TryStatement;
             this.block = block;
-            this.guardedHandlers = guardedHandlers;
-            this.handlers = handlers;
+            this.guardedHandlers = [];
+            this.handlers = handler ? [ handler ] : [];
```
ikarienator: We can in turn add a getter printing obsoletion messages.

mikesherov: Yes, but perhaps we need to introduce backCompat mode for this. Because printing an obsolete warning is unexpected output. Unless you are saying print obsolete notice in 3.0.0 and remove in 4.0.0, but that seems a bit too cautious.

ariya: Giving a warning right now is too premature anyway. We shall try to adjust popular downstream project to track `handler` instead and see if further migration warrants giving the warning or not.

ikarienator:
Good point. This looks good to me!

mikesherov:
:+1: @ariya feel free to land this

PR #1036, opened 2015-02-14
Expand property method range and loc to include params

### LINKED ISSUE #1029: Range for object literal shorthand methods doesn't include params
**What steps will reproduce the problem?**

On a fresh checkout of the harmony branch:

```javascript
// test.js
var esprima = require('./');

var source = [
    '({',
    '    method1(arg1) {},',
    '})'
].join('\n');

var ast = esprima.parse(source, { range: true });

console.log(ast.body[0].expression.properties[0]);
console.log(ast.body[0].expression.properties[0].value.params[0]);
```

```
$ node test.js
{ type: 'Property',
  key: { type: 'Identifier', name: 'method1', range: [ 7, 14 ] },
  value:
   { type: 'FunctionExpression',
     id: null,
     params: [ [Object] ],
     defaults: [],
     body: { type: 'BlockStatement', body: [], range: [Object] },
     rest: null,
     generator: false,
     expression: false,
     range: [ 21, 23 ] },
  kind: 'init',
  method: true,
  shorthand: false,
  computed: false,
  range: [ 7, 23 ] }
{ type: 'Identifier', name: 'arg1', range: [ 15, 19 ] }
```

**What is the expected output?**

I would expect the FunctionExpression's range to include the method's parameters as well as the body as in `(arg1) {}`.

It makes sense that the method's name is not included, as that is the Property's key and does not belong to the FunctionExpression.

**What do you see instead?**

The FunctionExpression's range includes only the body, `{}`, leading to a scenario in which a traversal starting at the FunctionExpression, range 21-23, would reach the parameter, whose range, 15-19, lies outside that of its parent.

(Migrated from https://code.google.com/p/esprima/issues/detail?id=625, as reported by @btmills)

### LINKED ISSUE #1040: Location for property getting/setter should be extended
Consider the following test code:

```js
var esprima = require('./esprima');
var code = '({ set width(w) {} })';
var tree = esprima.parse(code, { loc: true });
console.log(tree.body[0].expression.properties[0].key.loc.start.column);
console.log(tree.body[0].expression.properties[0].value.loc.start.column);
console.log(tree.body[0].expression.properties[0].value.params[0].loc.start.column);
```

The current master branch produced the following output:

```
7
16
13
```

This is because the `key` of `Property` node looks like:

```js
{ loc: { start: { line: 1, column: 7 }, end: { line: 1, column: 12 } },
  type: 'Identifier',
  name: 'width' }
```

and the `value` of `Property` node:

```js
{ loc: { start: { line: 1, column: 16 }, end: { line: 1, column: 18 } },
  type: 'FunctionExpression',
  id: null,
  params: [ { loc: [Object], type: 'Identifier', name: 'w' } ],
  defaults: [],
  body: 
   { loc: { start: [Object], end: [Object] },
     type: 'BlockStatement',
     body: [] },
  rest: null,
  generator: false,
  expression: false }
```

with its first parameter as:

```js
{ loc: { start: { line: 1, column: 13 }, end: { line: 1, column: 14 } },
  type: 'Identifier',
  name: 'w' }
```

This is different than the result of SpiderMonkey (`Reflect.parse` instead of `esprima.parse`):

```
7
7
13
```

For the analysis, this is how the `Property` looks like:

```json
{
  "loc": {
    "start": {
      "line": 1,
      "column": 7
    },
    "end": {
      "line": 1,
      "column": 18
    },
    "source": null
  },
  "type": "Property",
  "key": {
    "loc": {
      "start": {
        "line": 1,
        "column": 7
      },
      "end": {
        "line": 1,
        "column": 12
      },
      "source": null
    },
    "type": "Identifier",
    "name": "width"
  },
  "value": {
    "loc": {
      "start": {
        "line": 1,
        "column": 7
      },
      "end": {
        "line": 1,
        "column": 18
      },
      "source": null
    },
    "type": "FunctionExpression",
    "id": null,
    "params": [
      {
        "loc": {
          "start": {
            "line": 1,
            "column": 13
          },
          "end": {
            "line": 1,
            "column": 14
          },
          "source": null
        },
        "type": "Identifier",
        "name": "w"
      }
    ],
    "defaults": [],
    "body": {
      "loc": {
        "start": {
          "line": 1,
          "column": 16
        },
        "end": {
          "line": 1,
          "column": 17
        },
        "source": null
      },
      "type": "BlockStatement",
      "body": []
    },
    "rest": null,
    "generator": false,
    "expression": false
  },
  "kind": "set",
  "method": false,
  "shorthand": false
}
```

The result from SpiderMonkey is more sensible. Note how the `value` of the Property node starts as the same place (`column: 7`) as the `key`.

btmills:
Previously, the range and loc for get, set, and shorthand methods inside object literal properties included only the method body, meaning a method's parameters were outside the range of their parent `FunctionExpression` node. This was happening because the `FunctionExpression` node wasn't being created until `parsePropertyFunction` was called, by which point the method's parameters had already been parsed. This PR solves the issue by creating the `FunctionExpression` node as soon as it determines that the property is a get, set, or shorthand method, then passing the node into `parsePropertyFunction`.

If this gets merged, I'll port the same change to Espree as the fix for eslint/espree#36, which will enable a solution for eslint/eslint#1680.

I attempted to follow the contribution guidelines, but please let me know if I missed anything.

Closes #1029
Closes #1040

### DIFF
--- a/esprima.js
+++ b/esprima.js
@@ -2219,8 +2219,8 @@
 
     // 11.1.5 Object Initialiser
 
-    function parsePropertyFunction(param, first) {
-        var previousStrict, body, node = new Node();
+    function parsePropertyFunction(node, param, first) {
+        var previousStrict, body;
 
         previousStrict = strict;
         body = parseFunctionSourceElements();
@@ -2232,12 +2232,12 @@
     }
 
     function parsePropertyMethodFunction() {
-        var previousStrict, param, method;
+        var previousStrict, param, method, node = new Node();
 
         previousStrict = strict;
         strict = true;
         param = parseParams();
-        method = parsePropertyFunction(param.params);
+        method = parsePropertyFunction(node, param.params);
         strict = previousStrict;
 
         return method;
@@ -2262,7 +2262,7 @@
     }
 
     function parseObjectProperty() {
-        var token, key, id, value, param, node = new Node();
+        var token, key, id, value, param, methodNode, node = new Node();
 
         token = lookahead;
 
@@ -2274,23 +2274,25 @@
 
             if (token.value === 'get' && !(match(':') || match('('))) {
                 key = parseObjectPropertyKey();
+                methodNode = new Node();
                 expect('(');
                 expect(')');
-                value = parsePropertyFunction([]);
+                value = parsePropertyFunction(methodNode, []);
                 return node.finishProperty('get', key, value, false, false);
             }
             if (token.value === 'set' && !(match(':') || match('('))) {
                 key = parseObjectPropertyKey();
+                methodNode = new Node();
                 expect('(');
                 token = lookahead;
                 if (token.type !== Token.Identifier) {
                     expect(')');
                     tolerateUnexpectedToken(token);
-                    value = parsePropertyFunction([]);
+                    value = parsePropertyFunction(methodNode, []);
                 } else {
                     param = [ parseVariableIdentifier() ];
                     expect(')');
-                    value = parsePropertyFunction(param, token);
+                    value = parsePropertyFunction(methodNode, param, token);
                 }
                 return node.finishProperty('set', key, value, false, false);
             }
--- a/test/test.js
+++ b/test/test.js
@@ -1535,9 +1535,9 @@ var testFixture = {
                             rest: null,
                             generator: false,
                             expression: false,
-                            range: [18, 36],
+                            range: [15, 36],
                             loc: {
-                                start: { line: 1, column: 18 },
+                                start: { line: 1, column: 15 },
                                 end: { line: 1, column: 36 }
                             }
                         },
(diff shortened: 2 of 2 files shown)


ikarienator:
This looks good to me :+1:

btmills:
This change applies to the shorthand methods from #1029 and to `get` and `set` methods from #1040. However, it differs from SpiderMonkey's implementation, which overlaps the key with the `FunctionExpression` node; this PR has the `FunctionExpression` start immediately after the key, at the opening paren for the function's params, as @ikarienator described. I've amended the commit message to append "Closes #1040" so this can be merged immediately if that's acceptable. Otherwise I can investigate how to make the nodes overlap.

ariya:
@btmills Can you sign the [CLA](http://contribute.jquery.org/CLA/)? Be sure to use the same name and email as you used for the Github commit.

btmills:
Done

ariya:
Landed. Thank you @btmills!

PR #1039, opened 2015-02-15
fixes #1037: ES6: support computed property names

### LINKED ISSUE #1037: ES6 feature: computed property name
## Syntax:
<pre>
<i>ComputedPropertyName</i>[Yield] :
    <b>[</b> <i>AssignmentExpression</i><sub>[In, ?Yield]</sub> <b>]</b>
</pre>

**Referred to by:**
<pre>
<i>PropertyName</i>[Yield,GeneratorParameter] :
    <i>LiteralPropertyName</i>
    [+GeneratorParameter] <i>ComputedPropertyName</i>
    [~GeneratorParameter] <i>ComputedPropertyName</i><sub>[?Yield]</sub>
</pre>

## Early errors:

None.

## Spec:
[12.2.5](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-object-initializer) Object Initializer

## See also
[Esprima issue on Google Code](https://code.google.com/p/esprima/issues/detail?id=552)
[V8 support for computed property names](https://code.google.com/p/v8/issues/detail?id=3754)
[Error messages in V8](https://github.com/v8/v8-git-mirror/blob/4.2.70/src/messages.js)

ikarienator:
Support for computed property names (#1037).

This also relaxes the restrictions on duplicate property names according to the new ES6 spec. It is a precursor for this change.

I'll fix the AST format after the discussion on the topic.
I will also be happy to rebase after #1036.


test/reflect.js
```diff
@@ -100,7 +100,7 @@ function newExpr(callee, args) { return Pattern({ type: "NewExpression", callee:
 function callExpr(callee, args) { return Pattern({ type: "CallExpression", callee: callee, arguments: args }); }
 function arrExpr(elts) { return Pattern({ type: "ArrayExpression", elements: elts }); }
 function objExpr(elts) { return Pattern({ type: "ObjectExpression", properties: elts }); }
-function objProp(key, value, kind) { return Pattern({ type: "Property", key: key, value: value, kind: kind, method: false, shorthand: false }); }
+function objProp(key, computed, value, kind) { return Pattern({ type: "Property", key: key, computed: computed, value: value, kind: kind, method: false, shorthand: false }); }
```
ariya: It is better to hardcode `computed` to `false` here and keep the rest of the file unmodified.

esprima.js
```diff
@@ -192,13 +185,11 @@
         StrictFunctionName: 'Function name may not be eval or arguments in strict mode',
         StrictOctalLiteral: 'Octal literals are not allowed in strict mode.',
         StrictDelete: 'Delete of an unqualified identifier in strict mode.',
-        StrictDuplicateProperty: 'Duplicate data property in object literal not allowed in strict mode',
-        AccessorDataProperty: 'Object literal may not have data and accessor property with the same name',
-        AccessorGetSet: 'Object literal may not have multiple get/set accessors with the same name',
```
ariya: If these 3 checks are gone, we need to note the loss of these ES5-specific features (if one still wants to validate ES5-only code).

esprima.js
```diff
@@ -192,13 +185,11 @@
...
         StrictLHSAssignment: 'Assignment to eval or arguments is not allowed in strict mode',
         StrictLHSPostfix: 'Postfix increment/decrement may not have eval or arguments operand in strict mode',
         StrictLHSPrefix: 'Prefix increment/decrement may not have eval or arguments operand in strict mode',
-        StrictReservedWord: 'Use of future reserved word in strict mode'
+        StrictReservedWord: 'Use of future reserved word in strict mode',
+        DuplicateProtoProperty: 'Duplicate __proto__ fields are not allowed in object literals'
```
ariya: Could you confirm that this is the error message that V8 is using?

sebmck: [It does appear to be](https://github.com/v8/v8-git-mirror/blob/dff690ec0763721808582c7bd382768091648d0e/src/messages.js#L182).

ikarienator: Yes. I don't usually put English in my code. But when I do, I copy them from other people. :ghost:

esprima.js
ikarienator: Sure, how should we do it? Although I think it's ***WRONG*** to check early errors in ES5-only code using an ES6 parser.

test/test.js
```diff
@@ -29565,12 +30092,7 @@ var testFixture = {
                 start: { line: 1, column: 0 },
                 end: { line: 1, column: 48 }
             },
-            errors: [{
-                index: 46,
-                lineNumber: 1,
-                column: 47,
-                message: 'Error: Line 1: Object literal may not have multiple get/set accessors with the same name'
-            }]
+            errors: []
```
ikarienator: @ariya Since this is not erroneous, should I remove them completely?

ariya: Most likely, it needs to be moved out of the tolerant testing as well.

esprima.js
ariya: Not sure how. We just need to note it somewhere. Resolving the ES5 vs ES6 needs to happen somewhere else.

test/test.js
ikarienator: Done.

esprima.js
```diff
@@ -2293,74 +2306,74 @@
...
             }
-            if (match('(')) {
-                value = parsePropertyMethodFunction();
-                return node.finishProperty('init', id, value, true, false);
+            // finished
+            return methodOrPropertyName;
+
+        default:
+            // init property or short hand property.
+            if (isProto(methodOrPropertyName[0], methodOrPropertyName[1])) {
```
ariya: I find this is a bit confusing, consider the previous call (for `Syntax.Property`) is using `key` and `computed` properties as the parameters, not array indexes.

esprima.js
```diff
@@ -2293,74 +2306,74 @@
...
-                        tolerateError(Messages.AccessorGetSet);
-                    }
-                }
-                map[key] |= kind;
-            } else {
-                map[key] = kind;
-            }
-
-            properties.push(property);
+            properties.push(parseObjectProperty(hasProto));
```
ariya: I think this is an opportunity to experiment with that style of auxiliary pass to determine an additional error (in this case, duplicated proto). It seems that the code is cleaner with another loop than passing around the `hasProto` state.
It may not need be in this patch (could be a subsequent clean-up), though.

test/reflect.js
```diff
@@ -407,13 +407,9 @@ assertStmt("function f() { var x = 42; var x = 43; }",
 // getters and setters
 
  assertExpr("({ get x() { return 42 } })",
-            objExpr([ objProp(ident("x"),
-                              funExpr(null, [], blockStmt([returnStmt(lit(42))])),
-                              "get" ) ]));
+            objExpr([ objProp(ident("x"), funExpr(null, [], blockStmt([returnStmt(lit(42))])), "get") ]));
```
ariya: whitespace changes only? Better leave it as is, to avoid git blame confusion.

test/reflect.js
```diff
@@ -407,13 +407,9 @@ assertStmt("function f() { var x = 42; var x = 43; }",
...
  assertExpr("({ set x(v) { return 42 } })",
-            objExpr([ objProp(ident("x"),
-                              funExpr(null, [ident("v")], blockStmt([returnStmt(lit(42))])),
-                              "set" ) ]));
+            objExpr([ objProp(ident("x"), funExpr(null, [ident("v")], blockStmt([returnStmt(lit(42))])), "set") ]));
```
ariya: ditto.

esprima.js
ikarienator: The main difficulty right now is to provide the accurate location information especially when the tree does not carry it. Let's discuss this on Wednesday's meeting.

ikarienator:
Bother... Github stopped to diff the long test.js file... Another reason to break it down!

ariya:
Looks good now :+1: Can you also rebase @ikarienator?

esprima.js
mikesherov: Yeah, we may want to have es5 vs. es6 mode. I know a few downstream projects like ESLint and JSCS want to be able to still error on es5 code that is non conformant.

esprima.js
mikesherov: Noted in the agenda.

ikarienator:
Rebased.

ikarienator:
Rebased again.

ariya:
Landed. Thank you @ikarienator

PR #1050, opened 2015-02-16
Introduce rest parameters for Function Expressions and Declarations

mikesherov:
Also adds missing devDependencies that are required for a full `npm test` run.

### DIFF
--- a/esprima.js
+++ b/esprima.js
@@ -198,7 +198,10 @@
         StrictLHSAssignment: 'Assignment to eval or arguments is not allowed in strict mode',
         StrictLHSPostfix: 'Postfix increment/decrement may not have eval or arguments operand in strict mode',
         StrictLHSPrefix: 'Prefix increment/decrement may not have eval or arguments operand in strict mode',
-        StrictReservedWord: 'Use of future reserved word in strict mode'
+        StrictReservedWord: 'Use of future reserved word in strict mode',
+        ParameterAfterRestParameter: 'Rest parameter must be last formal parameter',
+        DefaultRestParameter: 'Unexpected token =',
+        ObjectPatternAsRestParameter: 'Unexpected token {'
     };
 
     // See also tools/generate-unicode-regex.py.
@@ -674,6 +677,19 @@
 
         // Check for most common single-character punctuators.
         case 0x2E:  // . dot
+            ch3 = source.substr(index, 3);
+            if (ch3 === '...') {
+                index += 3;
+                return {
+                    type: Token.Punctuator,
+                    value: ch3,
+                    lineNumber: lineNumber,
+                    lineStart: lineStart,
+                    start: start,
+                    end: index
+                };
+            }
+            /*falls through*/
         case 0x28:  // ( open bracket
         case 0x29:  // ) close bracket
         case 0x3B:  // ; semicolon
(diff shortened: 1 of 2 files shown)


mikesherov:
@ariya I skipped doing arrow function rest parameters because that's currently still over my head, requiring changes to `parseExpression` and adding `parseSpreadOrAssignmentExpression` support. Perhaps @jeffmo or someone a bit more up to speed with the codebase can handle that portion of it?

ikarienator:
We just found out a simple way to properly handle ArrowFunctionExpression
that only involves parseGroup and parseAssignmentExpression, we should port
it to esprima.

ariya:
@mikesherov Hmm, which part of `npm test` fails for you if `estraverse` and `esutils` are not in package.json's devDependencies?

ariya:
@ikarienator Sounds good. Please file a new enhancement issue.

esprima.js
```diff
@@ -673,7 +676,6 @@
         switch (code) {
 
         // Check for most common single-character punctuators.
-        case 0x2E:  // . dot
```
ikarienator: I would try to do it here since it's the most common punctuator.

mikesherov: Good call.

ikarienator:
Created: #1056

mikesherov:
@ariya, running `npm test` fails when it gets to doing coverage analysis.

ariya:
@mikesherov Can you double check? The build jobs at Travis CI never showed any error with package.json as it is now.

mikesherov:
@ariya ok, I'll back it out. Must be something weird about my local npm.

mikesherov:
@ariya @ikarienator ready for another round of review.

esprima.js
```diff
@@ -674,6 +677,19 @@
 
         // Check for most common single-character punctuators.
         case 0x2E:  // . dot
+            ch3 = source.substr(index, 3);
```
ariya: While this might look like a micro-optimization, I think we should consider checking for another dot before making `substr` call. For example, in a library as large as jQuery, there could > 5000 hits to this place. Since it follows that `...` is not widely used, it is safe to assume that we need to fail fast in that case, without incurring the function call overhead.

Also, it seems like an arbitrary choice, but I think falling through here it's not necessary (i.e. just return the dot token). I am ambivalent about each variant, although it's usually safe to err on the not-falling-through case.

Having said that, I think it's fine to address that minor tweak in a separate follow-up.

esprima.js
```diff
@@ -198,7 +198,10 @@
         StrictLHSAssignment: 'Assignment to eval or arguments is not allowed in strict mode',
         StrictLHSPostfix: 'Postfix increment/decrement may not have eval or arguments operand in strict mode',
         StrictLHSPrefix: 'Prefix increment/decrement may not have eval or arguments operand in strict mode',
-        StrictReservedWord: 'Use of future reserved word in strict mode'
+        StrictReservedWord: 'Use of future reserved word in strict mode',
+        ParameterAfterRestParameter: 'Rest parameter must be final parameter of an argument list',
+        DefaultRestParameter: 'Rest parameter can not have a default value',
+        ObjectPatternAsRestParameter: 'Invalid rest parameter'
```
ariya: As usual, let's check if V8 already supplies such an error message.

mikesherov: How do I go about checking these?

ariya: https://chromium.googlesource.com/v8/v8.git/+/master/src/messages.js is a good source, beside the usual hassle of checking out and building V8.

ariya:
LGTM :+1:

esprima.js
mikesherov: There seems to be no error message for this nor defaultRestParameter. I suppose leave those as is? I'll fix the other one though as the message is wrong.

esprima.js
mikesherov: It's either substr, or charCodeAt followed by potentially substr. Is charCodeAt that much faster that it's worth doing?

ikarienator: In theory, `charCodeAt` can be much faster but probably vms are not smart enough to optimize it. As I tested, there is no significant differences in numbers.

mikesherov: OK, so then going to leave as is for now.

mikesherov:
OK, so I updated the 1 error message and filed https://github.com/jquery/esprima/issues/1062 for the others. @ariya please land this when you're ready. I've already signed the CLA (and in fact, a CAA for all jquery projects).

mikesherov:
As discussed on #1062, I updated the error messages to `unexpected token` for the non standard error message cases. Going to land this now.

Issue #1065, opened 2015-02-18
ES6 feature: Let and Const Declaration


ariya:
## Syntax:

Per Rev 32 (Feb 2, 2015) of the draft spec, the relevant grammar is on [Section 13.2.1](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-let-and-const-declarations) on *Let and Const Declarations*:

```
LexicalDeclaration[In, Yield] :
  LetOrConst BindingList[?In, ?Yield] ;

LetOrConst :
  let
  const

BindingList[In, Yield] :
  LexicalBinding[?In, ?Yield]
  BindingList[?In, ?Yield] , LexicalBinding[?In, ?Yield]

LexicalBinding[In, Yield] :
  BindingIdentifier[?Yield] Initializer[?In, ?Yield]opt
  BindingPattern[?Yield] Initializer[?In, ?Yield]
```

**Referred to by:**

Declaration:

```
Declaration[Yield] :
  HoistableDeclaration _[?Yield]
  ClassDeclaration[?Yield]
  LexicalDeclaration[In, ?Yield]
```

StatementListItem:

```
StatementListItem[Yield, Return] :
  Statement[?Yield, ?Return]
  Declaration[?Yield]
```

Iteration:

```
for ( LexicalDeclaration[?Yield] Expression[In, ?Yield]opt ; Expression[In, ?Yield]opt ) Statement[?Yield, ?Return]
for ( ForDeclaration[?Yield] in Expression[In, ?Yield] ) Statement[?Yield, ?Return]
```

whereas:

```
ForDeclaration[Yield] :
  LetOrConst ForBinding[?Yield]

ForBinding[Yield] :
  BindingIdentifier[?Yield]
  BindingPattern[?Yield]
```

## Early errors:

* [x] const without initializer: `const x;` (from Esprima 1.x)
* [x] for-in with multiple let identifiers: `for (let x,y in z){}`
* [x] for-in with multiple const identifiers: `for (const x,y in z){}`
* [x] for-in with let initializer: `for (let i = 1 in x){}`
* [x] for-in with const initializer: `for (const i = 1 in x){}`
 * More TBD

## Spec:

[13.2.1](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-let-and-const-declarations) *Let and Const Declarations*.

Relevant AST interfaces (from https://github.com/estree/estree):

```js
interface VariableDeclaration <: Declaration {
    type: "VariableDeclaration";
[...]

PR #1070, opened 2015-02-19
Support for ES6 lexical declaration in switch statement.

ariya:
Refs #1065.

### DIFF
--- a/esprima.js
+++ b/esprima.js
@@ -3420,7 +3420,7 @@
             if (match('}') || matchKeyword('default') || matchKeyword('case')) {
                 break;
             }
-            statement = parseStatement();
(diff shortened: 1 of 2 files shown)


ariya:
@ikarienator Please review. This is just the first step since I can't do much on refactoring `parseSourceElements` until the class implementation is landed (unless you're fine with rebasing your class implementation later).

ikarienator:
Should we check duplicated symbols? It's also an Early Error to have duplicated bound names across case clauses.
Otherwise :+1:.

ariya:
@ikarienator Sure, more is coming in the subsequent lexical declaration refactoring.

PR #1075, opened 2015-02-22
Unit Tests: break into much smaller files

### LINKED ISSUE #1025: Unit Tests: break into much smaller files
We should take a page from espree here and break out the test files into much smaller pieces organized by the main language feature they are testing.

Right now it is unwieldily working with the giant file, and intimidating to newer contributors, like myself.

ikarienator:
Splits `test.js` file into small files, with "`test/fixtures/{Category}/{Test Name}.{Resource Kind}.(js|json)`" naming convention. 4 <i>Resource Kind</i>s are used:

0. `tree`: a JSON file representing the result of a parsing result. RegExps are stringified as `{}`.
0. `tokens`: a JSON file representing the result of a tokenization result.
0. `failure`: a JSON file representing an error as defined in the old `test/test.js` file. Field `tokenize` indicates whether the tests should invoke tokenizer instead of parser.
0. `result`: a JSON file representing the result of an API test.
0. `case`: consists the original source code string of the test.js test case.
0. `source`: an executable ES5 program defining a `source` variable in the global scope. Useful for unicode tests since escapes are visible.
0. `run`: an executable ES5 program. Useful for API tests. The code will be `eval`ed and the result value is used to compare against the expected result of the test case.

The migration simply names each test case with `migration_` followed by a number unique within the same folder.

Closes #1025.


michaelficarra:
Instead of sticking things on the global, why didn't you stick them on `module.exports`?

ikarienator:
It is harder to do it on the browser without using compiling tools. This
allow you to load the cases through xhr and eval it in the browser.

ariya:
Excellent work @ikarienator! I have few nitpicky questions, but I think overall this is an awesome step in the right direction.

ariya:
First of all, I don't think you need to worry much right now on running the test through the browser. Once we have a way to utilize a test runner such as Karma or Venus.js, it's the responsibility of the test runner to assemble everything to carry out the test.

Thus, instead of `sample.result.js` (that starts with `result = {...`), we can just `sample.result.json` (which is the straight JSON representation of the parsing result). Also, I don't have any strong opinion, I think having `syntax` and `tokens` as two different *Resource Kinds* (instead of just `result`) will facilitate easy testing of both `parse` and `tokenize`, respectively.

package.json
```diff
@@ -65,7 +65,7 @@
 
         "lint": "npm run check-version && npm run eslint && npm run jscs && npm run complexity",
         "check-version": "node tools/check-version.js",
-        "jscs": "jscs esprima.js test/*test.js",
+        "jscs": "jscs esprima.js && find test/fixtures -name *.result.js | xargs jscs",
```
ariya: I think you can use glob pattern here, e.g. `jscs  test/fixtures/**/*.result.js`. @mikesherov Can you confirm?

mikesherov:
Agreed with splitting syntax and tokens. Need an easy way to not have to do 1-to-1 correspondence. For example, both modules and template literals have no tests for tokens, and have bugs. Adding token tests needs to be much easier.

I also agree that the manifest file seems like too much work. Browser tests will have to rely on the runner.

ariya:
Also, if the result is JSON and we separate `syntax` and `tokens`, I suggest that we don't keep to keep a specific `case` *Resource Kind* for the source code to be tested. The relation will be:

  * `foo.js` is `parse`-d and verify against `foo.syntax.json`
  * `foo.js` is `tokenize`-d and verify against `foo.tokens.json`

Thoughts?

package.json
mikesherov: Should be able to. Might need to be quoted though depending on OS.

ariya:
```diff
diff --git a/test/runner.js b/test/runner.js
index 6811f65..29808a0 100644
--- a/test/runner.js
+++ b/test/runner.js
@@ -76,7 +76,7 @@ function sortedObject(o) {
     if (o === null) {
         return o;
     }
-    if (o instanceof Array) {
+    if (Array.isArray(o)) {
         return o.map(sortedObject);
     }
     if (typeof o !== 'object') {
```

IIRC `Array.isArray` is not supported on IE 8. If we want to use it, a simple polyfill is needed. We can also worry about this once we face the problem of running the tests in a web browser.

mikesherov:
@ariya, makes sense to me, except that forces 1-to-1 correspondence between token and syntax tests. Maybe caveat is: only tokenize if a foo.tokens.json exists?

mikesherov:
Yeah, ES5 shim is required in that case.

ariya:
Also, for the case enumeration, maybe should a two-digit or three-digit numbering (perhaps even starting from 1). So, instead of `0.js` to `19.js`, we can have `01.js` to `19.js`. This makes it looks nice, in appearance and when sorted, when doing `ls`, `find`, or just `git show`.

mikesherov:
> Also, for the case enumeration, maybe should a two-digit or three-digit numbering (perhaps even starting from 1). So, instead of 0.js to 19.js, we can have 01.js to 19.js.

I would imagine in the future we want real descriptive names, and that this was a brute force splitting.

ariya:
>
> @ariya <https://github.com/ariya>, makes sense to me, except that forces
> 1-to-1 correspondence between token and syntax tests. Maybe caveat is: only
> tokenize if a foo.tokens.json exists?
>
Definitely.

Now that you mention this, `error` might be also useful (but not necessary)
as another type. Makes it easy to track invalid syntax.

ariya:
>
> I would imagine in the future we want real descriptive names, and that
> this was a brute force splitting.
>
True. Piecemeal renaming of the existing tests is also possible.

ikarienator:
> Thus, instead of sample.result.js (that starts with result = {...), we can just sample.result.json (which is the straight JSON representation of the parsing result). Also, I don't have any strong opinion, I think having syntax and tokens as two different Resource Kinds (instead of just result) will facilitate easy testing of both parse and tokenize, respectively.

The reason I failed to make them json is because the tree is not always serializable to json due to <i>RegExp</i>s. But I can make those serializable json files.

I'm not sure how test runner can enumerate directory on a web page. We need to run this on the `test/index.html` in the gh-pages branch. But we can think about that  later.

@michaelficarra Since we've decided to use a test runner, for those we cannot serialize, I can export using `module.exports` now.

> Also, for the case enumeration, maybe should a two-digit or three-digit numbering (perhaps even starting from 1). So, instead of 0.js to 19.js, we can have 01.js to 19.js. This makes it looks nice, in appearance and when sorted, when doing ls, find, or just git show.

@ariya I think I shoule name them `migrated_####` (always 4 digits) so future tests won't be named using the numbers.

Question about the result file name. We have another type: "first in body". I think maybe it's time to always start from `Program` in parsing test?

BTW, the current explosion script: https://gist.github.com/ikarienator/8477a65559e471c73668.

## Tasks
- [ ] Rename _Test Name_ to `migrated_####`.
- [ ] Remake serializable trees to JSON.
- [ ] Export non serializable trees using `module.exports`.
- [ ] Separate parse tests and token tests.

package.json
ikarienator: Ah. I should always quote it since I don't want expansion.

nzakas:
It's not a good idea to test tokens only with `tokenize()`, since `parse()` follows a slightly different execution path. It's possible for `tokenize()` to pass and `parse()` to fail (we ran into this in Espree). I'd strongly recommend testing tokens with both methods.

Also, the AST doesn't convert cleanly into JSON due to the use of `undefined` and regex literals. In Espree, we made them JS files so we could go in and manually fix those issues.

ikarienator:
@nzakas Where is `undefined` used?

ariya:
>
> The reason I failed to make them json is because the tree is not always
> serializable to json due to *RegExp*s. But I can make those serializable
> json files.
>
It might be a bit cumbersome, but can you also have foo.regex (plain text
file) for the array of collected regex (from the raw property of each)? We
can test each regex via its pattern and flags already, both are string and
serializable.

@ariya <https://github.com/ariya> I think I shoule name them migrated_####
> (always 4 digits) so future tests won't be named using the numbers.
>
Good call.


> Question about the result file name. We have another type: "first in
> body". I think maybe it's time to always start from Program in parsing
> test?
>
Sure, go ahead with it.

>

ariya:
> It's not a good idea to test tokens only with tokenize(), since parse() follows a slightly different execution path. It's possible for tokenize() to pass and parse() to fail (we ran into this in Espree). I'd strongly recommend testing tokens with both methods.

Sure. Nothing above implies that running `parse()` with `{ tokens:true }` option won't be tested. Note that in this case, the result (the `tokens` array) is part of `.syntax.json`.

ikarienator:
@ariya 
> It might be a bit cumbersome, but can you also have foo.regex (plain text
file) for the array of collected regex (from the raw property of each)? We
can test each regex via its pattern and flags already, both are string and
serializable.

Not sure what it means? The AST tree for regular expression literals contains a regular expression which prevents serialization, what should we do with the AST tree?

ariya:
> Not sure what it means? The AST trees for regular expression literals contains regular expressions which prevents serialization, what should we do with the AST tree?

I'll come up with a more concrete example. In the mean time, I suggest we ignore that particular regex `value` (which JSON.stringified into `{}` anyway) and rely on testing its `raw`, `pattern` and `flags`.

ikarienator:
Hmm. "RegExp.prototype.toString" works differently in iojs...

mikesherov:
> Hmm. "RegExp.prototype.toString" works differently in iojs...

@ikarienator @ariya so is this mergeable now?

ariya:
We need to agree on the filename patterns. Right now, it's still:

  * migrated_0000.case.js
  * migrated_0000.result.json
  * migrated_0000.tokens.json (in some cases)

ikarienator:
I thought that's what we agreed on, no?

ikarienator:
@mikesherov This is currently mergeable.

ariya:
What I recommend:

  * migrated_0000.js (`case` is redundant information)
  * migrated_0000.syntax.json (`result` is too implicit)

At some point I imagine we can fold `source` and `run` in a different way, but I think it's fine to leave it as it.

michaelficarra:
Instead of syntax, I think we should use estree. Just in case esprima uses a different format one day...

ariya:
> Instead of syntax, I think we should use estree.

I was also thinking about it. I think it only works if Esprima output aligns completely with that spec. Any deviations or extensions will potentially cause a confusion.

ikarienator:
@michaelficarra How about just `tree`.

ariya:
> How about just tree.

I'm fine with `tree`.

ikarienator:
It's done.

ariya:
LGTM :+1:

ariya:
Spoke to @ikarienator, I think we both agree on removing `case` for simplicity.

mikesherov:
:+1:

ikarienator:
The `case` is removed.

ariya:
Landed with slightly modified commit message (no `case`). Thank you very much @ikarienator!

mikesherov:
@ariya, just a bit of housekeeping, the commit message should have been amended to include `Closes #1075` as well so the PR would be auto closed and we'd have reference to this discussion. Easy to miss, but just wanted to remention it.

mikesherov:
P.S. Super excited for this change!

fkling:
:+1: very nice

ikarienator:
BTW, for those who's working on `test/test.js`, you can use [this script](https://gist.github.com/ikarienator/8477a65559e471c73668) to migrate it to the new format.

nzakas:
:100:

PR #1076, opened 2015-02-22
More lexical declaration support

ariya:


mikesherov:
:+1: although should probably wait for the unit test split to land.

esprima.js
```diff
@@ -2974,13 +2982,29 @@
 
     // 12.1 Block
 
+    function parseStatementListItem() {
+        if (lookahead.type === Token.Keyword) {
+            switch (lookahead.value) {
+            case 'const':
+            case 'let':
```
ikarienator: Note that `let` will soon be an `Identifier`.

ariya: I plan to tackle that separately.

ariya:
Rebased against the latest master.

ariya:
@mikesherov Do I need to explicitly sign the CLA (even after the CAA and such)?

ikarienator:
I think we should use some meaningful names for test cases in the future. Otherwise LGTM :+1: !

mikesherov:
@ariya, sign the CLA just so the automated check passes.

scottgonzalez:
I'm curious what the terms of the CAA are and how it affects our checks.

PR #1091, opened 2015-03-05
\r doesn't increase line number

gabelevi:
The carriage return (aka `CR`, `13`, or `\r`) is a line terminator but should not increment the line number.


michaelficarra:
Carriage return without following line feed is still a line terminator and should increment the line number.

gabelevi:
I don't see line numbers specified in the ES6 spec. Why do you say CR should increment line numbers? This is not what I observe in the wild. For example

```Bash
[glevi@dev1475 esprima{terminator}]$ node -e "console.log('/*\r*/]')" > foo.js; node foo.js 

/data/users/glevi/esprima-glevi/foo.js:1
*/]nction (exports, require, module, __filename, __dirname) { /*
                                                                   ^
SyntaxError: Unexpected token ]
    at Module._compile (module.js:439:25)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
    at Function.Module.runMain (module.js:497:10)
    at startup (node.js:119:16)
    at node.js:906:3
[glevi@dev1475 esprima{terminator}]$ node -e "console.log('/*\n*/]')" > foo.js; node foo.js 

/data/users/glevi/esprima-glevi/foo.js:2
*/]
  ^
SyntaxError: Unexpected token ]
    at Module._compile (module.js:439:25)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
    at Function.Module.runMain (module.js:497:10)
    at startup (node.js:119:16)
    at node.js:906:3
```

In this example `\n` increments the line number while `\r` does not.

jeffmo:
As I understand it, it's not uncommon for text files to end with `\r\n` because most text editors deal only with \r, \n, or \r\n as a newline character (but not both at the same time).

I don't think we want \r\n to count as two increments to the line number from the parser as this is almost never the intention nor is it the behavior of most other things that deal with line breaks.

michaelficarra:
The spec doesn't say anything about line numbers, but it does specify line terminators, which I use to imply separation of lines (aka increased line numbers).

jeffmo:
(Specifically: Many windows systems use \r\n to indicate a single line break)

fkling:
> Carriage return without following line feed is still a line terminator and should increment the line number.

Yep, the spec clearly lists a `\r` as a line terminator. 

> I don't think we want \r\n to count as two increments to the line number from the parser as this is almost never the intention nor is it the behavior of most other things that deal with line breaks.

That is true as well and also specified in the spec:

>  The sequence `<CR><LF>` is commonly used as a line terminator. It should be considered a single SourceCharacter for the purpose of reporting line numbers.


https://people.mozilla.org/~jorendorff/es6-draft.html#sec-line-terminators

gabelevi:
To be 100% clear, we're all agreed that `<CR>` is a line terminator. Indeed, the first thing I said was

> The carriage return (aka `CR`, `13`, or `\r`) is a line terminator

The thing I'm arguing is that `CR` by itself shouldn't increment the line number (`<CR><LF>` clearly should). This is consistent with V8 and Unix and modern MacOS. Yeah, MacOS before OSX treated `<CR>` as a newline, but I don't see why we'd follow an old OS. Historically, esprima has tried to follow V8 for error messages.

@jeffmo mentioned that my node example might be demonstrating Unix's bias rather than V8. However if in your chrome browser console you do `eval("\r]")`, chrome will report the line number of the error as `1`, while it will say `2` if you use `\n`.

fkling:
> Historically, esprima has tried to follow V8 for error messages.
Can't argue against that.

FWIW, Firefox reports the error in line 2.

gabelevi:
I've signed the CLA 3 times...hopefully one of them will work.

michaelficarra:
Sounds like a V8 bug to me. No reason to copy it. Esprima follows the messaging, not necessarily the position of errors.

ariya:
Can someone file a ticket for this? I feel that the conversation (e.g. ref to spec, V8 behavior, etc) is better handled in the ticket.

jeffmo:
@gabelevi: I don't think github knows to re-run the CLA check unless you "touch" the diff somehow -- so if you've signed it, you're probably fine.

(If you really want the red stuff to go away, maybe add some whitespace to your diff, send it, delete the whitespace, then re-send just to probe things)

jeffmo:
@ariya : Seems a little easier to keep the discussion all in one place if possible -- especially since there's already ongoing discussion.

There's definitely room for discussion here since it seems v8/chrome and ff/spidermonkey do different things here... I guess we just have to pick which one we want to follow. At this point I think @gabelevi makes a fair point that almost no systems (based on some quick googling) actually seem to break the line on <CR> these days -- clearly with Firefox as one exception.

@michaelficarra: It sounds like you use CR -- can you talk more about some of the systems you use that follow this? It would be good to understand who we'd be misaligned with if we decided to take this.

ariya:
@jeffmo In general or for this one in particular? In all cases, it diverges from the project practices so far and hence I propose it as an agenda item for the next meeting.

jeffmo:
@ariya: At least this one -- although it'd be a shame to dissuade people from writing code for proposals up front even if the proposal might not go through. In this case the doability is obvious, but in many cases there's doubt or wonder about what it would look like if done. Having the change right in front of you often makes it much easier to decide if it's worth moving forward (code wins arguments, etc). I don't want to hijack this thread about processes, but I also don't want us to become slaves to them.

On waiting until the meeting next week -- it'd also be a shame to make @gabelevi have to wait a whole week for such a minor change. Can't we just work this through and get consensus from everyone here?

mikesherov:
@jeffmo, @ariya is saying make the concept of "is a ticket required" an agenda item.... not discussing this.

jeffmo:
Ahh ok, that makes sense -- sorry for the misunderstanding

ikarienator:
https://people.mozilla.org/~jorendorff/es6-draft.html#sec-line-terminators

> The sequence <CR><LF> is commonly used as a line terminator. It should be considered a single SourceCharacter for the purpose of reporting line numbers.

Thus :-1:.

jeffmo:
In the traditional sense, it seems <CR> returns the cursor back to the start of the line (but does not feed to the next line). This is useful for things like clearing a line on the console without scrolling existing output. LF is what actually moves to the next line. 

The spec only talks about line terminators, but it says nothing about line numbers -- so I think that's where Gabe is coming from here.

gabelevi:
@ikarienator: I think that line is just noting that `<CR><LF>` is a single line terminator sequence, as expressed in the grammar below. 

How should we make a decision here? I think the order of precedence should be 

1. Follow the ECMA spec
2. Follow other implementations
3. Follow common sense.

## The spec
My impression is that the ES6 spec doesn't specify line numbering. I think consensus is that line numbering is unspecified, but if some people disagree (sounds like @ikarienator disagrees) we could reach out to TC39 to clarify. If they choose to specify it we'll just do whatever they say. Otherwise, we have to move on to...

## Other implementations
I believe that historically esprima follows V8 on unspecified things (like error messages), though I believe this policy is changing. We do see Chrome and Firefox showing different behaviors here, so there definitely isn't consensus in other implementations on the right behavior. I would argue that we could also look at Windows, Unix, and OSX for support of a `<CR>` that is only a carriage return without a line feed, but OS's aren't what we normally look to when making these decisions.

## Common sense
If the spec leaves it unspecified and if other implementations conflict, then I think we just need to use common sense. I think common sense is strongly in favor of a `<CR>` that doesn't increment the line number. [The carriage return](http://en.wikipedia.org/wiki/Carriage_return) is the mechanism that moves the cursor to the beginning of the line. Some old operating systems (like MacOS before OSX) treat `<CR>` as a carriage return AND line feed, but modern OS's are pretty consistent at treating `<CR>` as a just a carriage return.

I should mention that this isn't the most important issue to me. This issue doesn't block me on anything.

mikesherov:
>  The carriage return is the mechanism that moves the cursor to the beginning of the line. Some old operating systems (like MacOS before OSX) treat <CR> as a carriage return AND line feed, but modern OS's are pretty consistent at treating <CR> as a just a carriage return.

Forgive my ignorance, honest question here. Wouldn't `<CR>` as **just** carriage return imply that `a<CR>b` means both `a` and `b` exist in column 1 of line 1?

ariya:
From http://stackoverflow.com/questions/12747722/what-is-the-difference-between-a-line-feed-and-a-carriage-return:

> A line feed means moving one line forward. The code is \n.
A carriage return means moving the cursor to the beginning of the line. The code is \r.

gibson042:
> Wouldn't `<CR>` as **just** carriage return imply that `a<CR>b` means both `a` and `b` exist in column 1 of line 1?

Yes, but "`a`" is overwritten (check for yourself in a terminal with `echo -e 'abc\r-'`).

The character's mechanics and history are not relevant here, however, because the ECMAScript lexical grammar dictates both of the following:
* `<CR>` code points are line terminators (thereby preventing potential collisions like the above)
* `<CR><LF>` terminates only one line, in contrast to every other pair of consecutive line terminators (which terminate *two* lines, the second of length 0).

michaelficarra:
> The character's mechanics and history are not relevant here

Agreed. This is what I've been trying to say.

jeffmo:
I think the point has been made many times already that specifying "line termination" in the spec doesn't say much about "line numbering" -- especially if real-world mechanics prove disagreement on this kind of interpretation (chrome vs ff).

If we choose not to take `\r` as a line incrementor, that's fine -- but I still think pervasive real-world mechanics should dictate what we do here. What systems (editors, OSes, etc) other than Firefox treat `\r` as a signal to increment line numbers or move the cursor to a new line?

mikesherov:
there are 3 choices:

1. CR increments line number.
2. CR does not increment line number AND cursor is set back to column 1.
3. CR does not increment line number ANS cursor is not set back to column 1.

Systems like consoles and editors implement choice 2. Javascript seems to imply choice 1. This PR seems to follow choice 3.

I'm not sure how you can read the specs language on line termination and not come to the conclusion of choice 1, but i don't personally see how choice 3 can be thought of as anything but a bug. 

What advantage do you gain by not incrementing line number after every line terminator other than confusion?

ariya:
Please note that V8 changed its behavior a few weeks ago, see https://github.com/v8/v8-git-mirror/commit/2050994d80:

> Correctly parse line ends for debugging.
Instead of using only \n as line terminator, we now use the definition
in http://www.ecma-international.org/ecma-262/5.1/#sec-7.3

Chrome 41 (stable) is still using V8 4.1 and hence it doesn't get the above changeset yet. Only V8 >= 4.2 will have the new behavior.

gabelevi:
> there are 3 choices:

>1. CR increments line number.
>2. CR does not increment line number AND cursor is set back to column 1.
>3. CR does not increment line number ANS cursor is not set back to column 1.


So how are line numbers used? A human might use line & column numbers to look up the corresponding piece of code. There are a bunch of ways to look up code.

node -e 'console.log("a\rb")' > foo.js

* `cat foo.js` - This uses (3) and looks like `b`
* `vim foo.js` - For me, this uses (2) and looks like `a^Mb`
* `emacs foo.js` - For me, this uses (2) and looks like `a^Mb`
* textedit & textmate - On my Mac, these both use (1)

How are line & column numbers used by non-humans? I'm not too sure. I bet @benjamn uses this stuff in [recast](https://github.com/benjamn/recast) and would probably prefer one of these behaviors.

ikarienator:
Please read this again:

> The sequence <CR><LF> is commonly used as a line terminator. It should be considered a single SourceCharacter for the purpose of reporting line numbers.

It basically implies that `<CR>` is  indeed a line terminator that increases line number, because otherwise this sentence doesn't need to be here. If you still think that is not clear enough, I filed a bug for ES6 spec here: https://bugs.ecmascript.org/show_bug.cgi?id=4141

------------------

The argument about how terminal show them is not relevant because `cat` doesn't specially process line terminators at all, its imply put it to the tty file. And different tty treats special control characters differently. The behavior that `\n` create a new line is basically incidental and this should not be an evidence on how you manage line number. For example, we have no reason to move column number when we see a `\b`; nor would we change the line number when we see a `\v`.

The argument about how editors show them is not too much relevant either because the editors you listed are all general purpose text editors. They have no reason to treat line terminators exactly like what a single language treats them, although ES5 does trying to comply to common usage of line terminators. Besides, vim and emacs show the `\r` as exactly a caret (`^`) and an `M`, which means they are none of (1), (2) or (3). Both emacs and vim will treat the `\r` to be two columns wide in their location indicators as well (for example, the `[` in `^M[`will be on col 3). I'm glad to see that modern GUI editors TextEdit and TextMate use (1). But to me it's still incidental. You may also find CodeMirror and Ace both use (1) as well.

It is wrong to use node to test this because the language node supports is too different from ES5/ES6. And since other implementations all move to recognize `\r` to be an actual "line terminator" as the spec implies, I think it's clear that we should comply to that as well.

ariya:
I think the most recent change in V8 should settle this. If I have a test file `test.js` containing `\r\r\r]`, then the most recent V8 shell (tested with 4.3.0) will give the following output:
```
test.js:4: SyntaxError: Unexpected token ]
]
^
SyntaxError: Unexpected token ]
```

mikesherov:
@gabelevi, thanks for sticking with us as we sorted out this issue. I agree with @ariya that the recent change in v8 puts this matter to bed. I suggest submitting a bug to TC39 if we want to disambiguate further. 

As far as this PR goes, I'm going to close for now. Feel free to reply further if you think this is the wrong course of action.

Thanks again!

PR #1092, opened 2015-03-05
Have let and const with initializer trigger a syntax error.

ariya:
Refs #1065


ariya:
Note that we use the generic "Unexpected token.." error message until V8 supports such an error and has a customized message.

ikarienator:
:+1:

Issue #1012, opened 2015-02-09
ES6 feature: Arrow Functions: tracking Issue


mikesherov:
## Syntax:

```
ArrowFunction[In, Yield] :
  ArrowParameters[?Yield] [no LineTerminator here] => ConciseBody[?In]
ArrowParameters[Yield] :
  BindingIdentifier[?Yield]
  CoverParenthesizedExpressionAndArrowParameterList[?Yield]
ConciseBody[In] :
  [lookahead ≠ { ] AssignmentExpression[?In]
  { FunctionBody }
```

## Supplemental Syntax

```
When the production 
  ArrowParameters[Yield] : CoverParenthesizedExpressionAndArrowParameterList[?Yield]
is recognized the following grammar is used to refine the interpretation of
  CoverParenthesizedExpressionAndArrowParameterList :
ArrowFormalParameters[Yield, GeneratorParameter] :
  ( StrictFormalParameters[?Yield, ?GeneratorParameter] )
```

## Spec:

https://people.mozilla.org/~jorendorff/es6-draft.html#sec-arrow-function-definitions

## Remaining Tasks:

 - [x] Review and Test Token Generation
 - [x] Review Error Messages
 - [x] Land Rest Parameters #1011 
 - [x] Fix #1009 
 - [x] Fix #1010 
 - [x] Fix #1013 
 - [x] Fix #1014 
 - [x] Fix #1056

Issue #1011, opened 2015-02-09
ES6 feature: Rest Parameters: tracking Issue


mikesherov:
## Syntax:

```
FormalParameterList[Yield,GeneratorParameter] :
  FunctionRestParameter[?Yield]
  FormalsList[?Yield, ?GeneratorParameter]
  FormalsList[?Yield, ?GeneratorParameter] , FunctionRestParameter[?Yield]
FunctionRestParameter[Yield] :
  BindingRestElement[?Yield]
BindingRestElement[Yield, GeneratorParameter] :
  [+GeneratorParameter] ... BindingIdentifier[Yield]
  [~GeneratorParameter] ... BindingIdentifier[?Yield]
```

## Spec:

https://people.mozilla.org/~jorendorff/es6-draft.html#sec-functions-and-classes

## Remaining Tasks:

 - [x] Review and Test Token Generation
 - [x] Review Error Messages
 - [x] Arrow Function Rest Parameters 
 - [ ] https://github.com/estree/estree/pull/37

mikesherov:
@ariya can you respond to the estree issue about rest and defaults? I think we may want to take acorns approach here.

ariya:
@mikesherov I'm not sure why we want to deviate from the existing SpiderMonkey AST for this case.

nzakas:
Just a heads up: I was doing some experimentation and found that there's no good way to support both the harmony branch `rest` property and the newly-agreed Acorn-like approach (having the rest param in the params array) at the same time. This is because code exists that loops over the params array and then checks to see if `rest` has a value. That means adding the rest param into the params array would result in double evaluation of a rest param.

So we can really only do one or the other but not both (as opposed to `handler` and `handlers`, for instance).

mikesherov:
@nzakas no way to add a flag rather than doing a second loop?

nzakas:
Sure, but that's still a breaking change. My point is that existing code will break if we support both at the same time. If we need a breaking change, might as well go all the way and remove `rest`.

mikesherov:
@nzakas I see, you mean consumer code. I'm +1 for removing all the way.

mikesherov:
@jeffmo ^ ?

jeffmo:
Agreed, let's rip the band-aid off

ariya:
Hmm, maybe I missed something, but when did I discuss keeping the compatibility? Rest parameter was not feature in 2.0 anyway.

jeffmo:
[...]

PR #1093, opened 2015-03-05
Support rest parameter in arrow expression

ikarienator:
Related to #1011, #1012.


ariya:
Typo: paramter -> parameter.

ariya:
Is the reoder of `rest` property in the node intentional or a side effect of the implementation?

ariya:
In `complex-rest-in-arrow-not-allowed.js`, there is no guarantee that the referred URL will be permanent. Might as well remove it or refer to the ES6 spec section.

ikarienator:
@ariya The change of order is intentional. `rest` parameter shows up in the syntax before body.

ikarienator:
The comment has been removed.

mikesherov:
Great work! Once this is done and landed, we need to do the work of replacing defaults/rest properties with what's outlined in espree/espree#3. I can take that if needed.

LGTM

mikesherov:
@ikarienator, can you make sure your commit message says:

```
Refs #1011
Refs #1012
```
?

Thanks!

ariya:
With the minor change and commit message refs to the tickets, it's :+1:

mikesherov:
:+1: from me, I'm going to land this.

Issue #1081, opened 2015-02-25
Use AssignmentExpression for default parameter value


ariya:
In estree/estree#3 and estree/estree#37, it was decided to replace `Function.defaults` with an array with `AssignmentPattern` as a formal parameter and remove the `Function.defaults` member. 

(This can not go into a release earlier than 3.0, because 2.0 was already released with support for defaults in the old fashion)

The relevant part of the AST format:

```js
interface Function <: Node {
    id: Identifier | null;
    params: [ Pattern ];
    body: BlockStatement;
}
```

where `Pattern` is either a `RestElement` (already implemented #1011) or `AssignmentExpression`.

As an example, the following code:
```js
esprima.parse('function f(x=42){}').body[0]
```
*currently* produces this output:
```js
FunctionDeclaration {
  type: 'FunctionDeclaration',
  id: Identifier { type: 'Identifier', name: 'f' },
  params: [ Identifier { type: 'Identifier', name: 'x' } ],
  defaults: [ Literal { type: 'Literal', value: 42, raw: '42' } ],
  body: BlockStatement { type: 'BlockStatement', body: [] },
  generator: false,
  expression: false }
```
*After* this feature will be implemented, the output would look like:
```js
FunctionDeclaration {
  type: 'FunctionDeclaration',
  id: Identifier { type: 'Identifier', name: 'f' },
  params: [ AssignmentPattern { type: 'AssignmentPattern', left: [Object], right: [Object] } ],
  body: BlockStatement { type: 'BlockStatement', body: [] },
  generator: false,
  expression: false }
```

ariya:
For the downstream test, we have to exclude js2coffee temporarily until this one is fixed: https://github.com/js2coffee/js2coffee/issues/427.

mikesherov:
Woah!

PR #1096, opened 2015-03-06
Introduce RestElement/AssignmentPattern as parameters instead of Function.rest and Function.defaults

mikesherov:


ariya:
I thought we want to at least preserve `defaults` (as a transition step, see #1081), so that it's not a breaking change until 3.0.

Consider the following test (with no rest parameters) at all:

```js
require('./esprima').parse('function f(x=0){}').body[0]
```

With 2.0, we will have the following output:

```js
{ type: 'FunctionDeclaration',
  id: { type: 'Identifier', name: 'f' },
  params: [ { type: 'Identifier', name: 'x' } ],
  defaults: 
   [ { type: 'Literal',
       value: 0,
       raw: '0' } ],
  body: { type: 'BlockStatement', body: [] },
  rest: null,
  generator: false,
  expression: false }
```

With the change in this PR, it is going to be:

```js
{ type: 'FunctionDeclaration',
  id: { type: 'Identifier', name: 'f' },
  params: 
   [ { type: 'AssignmentPattern',
       left: [Object],
       right: [Object] } ],
  body: { type: 'BlockStatement', body: [] },
  generator: false,
  expression: false }
```

mikesherov:
@ariya, meaning keep `defaults` as an empty array? or as an array of `undefined`? or keep the params as they were and make it additive?

mikesherov:
@ariya, oh right. Defaults were in 2.0. I'll back out defaults change, and make this just be RestElement. Then we can discuss strategy for defaults in #1081

mikesherov:
@ariya, I've updated the PR. @ikarienator @jeffmo can I get a review please? This is last blocker to 2.1

mikesherov:
I've signed the CLA, probably just a timing issue that @scottgonzalez is working through. I'll ensure the commit message says `Closes #1011` when I land it.

ariya:
LGTM :+1:

Issue #1089, opened 2015-03-04
ES6 feature: Update to ES6 identifier grammar


mathiasbynens:
Esprima currently implements ES5 identifier grammar. It would be nice to update to ES6.

In case it’s helpful — I just wrote a blog post that documents the differences between ES5 and ES6 identifiers, with examples: https://mathiasbynens.be/notes/javascript-identifiers-es6

ikarienator:
@mathiasbynens You might be interested in this as well https://bugs.ecmascript.org/show_bug.cgi?id=4097#c4

mathiasbynens:
@ikarienator Thanks for the heads up!

mathiasbynens:
@ikarienator Is this observable? How?

E.g. how should the following code behave:

```js
Function('var var;'); // grammar error → doesn’t throw (until executed)?
Function('var v\\u0061r;'); // early error → throws (before execution)?
```

michaelficarra:
@mathiasbynens They should both throw. I can't think of any way someone would be able to distinguish an early error from a grammar error.

ariya:
Since this is quite a big topic, I'll spawn few new tickets which will tackle the individual issues separately.

ariya:
This is now covered by all the most recent commits.

PR #1100, opened 2015-03-08
Make `testRegExp` less restrictive

mathiasbynens:
E.g.

```js
/[\uD834\uDF06-\uD834\uDF08a-z]/u
```

This is perfectly valid ES6 (given the `u` flag) but it didn’t parse in Esprima before this patch because `testRegExp` evaluates the regexp without the `u` flag (as it should, since it aims to support a wide range of browsers/environments).

This patch accounts for this case as well by using a regex to look for `\uXXXX` escape sequences and replacing them with a hardcoded ASCII letter `x` (like we previously did for `\u{…}` escapes).

Note: this still doesn’t handle all cases (but really, we’d need a regular expression parser to do that properly), e.g. when the high surrogate is represented by an escape sequence but the low surrogate is the raw character, but such cases are extremely rare to non-existent.

### DIFF
--- a/esprima.js
+++ b/esprima.js
@@ -1089,10 +1089,11 @@
         var tmp = pattern;
 
         if (flags.indexOf('u') >= 0) {
-            // Replace each astral symbol and every Unicode code point
-            // escape sequence with a single ASCII symbol to avoid throwing on
-            // regular expressions that are only valid in combination with the
-            // `/u` flag.
+            // Replace each astral symbol and every Unicode escape sequence
+            // that possibly represents an astral symbol or a paired surrogate
+            // with a single ASCII symbol to avoid throwing on regular
+            // expressions that are only valid in combination with the `/u`
+            // flag.
             // Note: replacing with the ASCII symbol `x` might cause false
             // negatives in unlikely scenarios. For example, `[\u{61}-b]` is a
             // perfectly valid pattern that is equivalent to `[a-b]`, but it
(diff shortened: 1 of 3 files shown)


mikesherov:
Thanks for contributing, @mathiasbynens! Can you please rename the added tests files to actually describe the test? `migrated_*` is a holdover from when we split up the giant monolith test file recently.

ariya:
Is this related to #1089?

mathiasbynens:
@mikesherov Done.

@ariya This is unrelated to #1089.

esprima.js
```diff
@@ -1089,10 +1089,11 @@
...
-            // regular expressions that are only valid in combination with the
-            // `/u` flag.
+            // Replace each astral symbol and every Unicode escape sequence
+            // that possibly represents an astral symbol or a paired surrogate
+            // with a single ASCII symbol to avoid throwing on regular
+            // expressions that are only valid in combination with the `/u`
+            // flag.
             // Note: replacing with the ASCII symbol `x` might cause false
             // negatives in unlikely scenarios. For example, `[\u{61}-b]` is a
             // perfectly valid pattern that is equivalent to `[a-b]`, but it
```
mikesherov: Is this note still valid?

mathiasbynens: Yes

ariya:
The bug was filed as #1114.

ariya:
LGTM.

PR #1104, opened 2015-03-10
Fix template literals token parsing by account for template context

mikesherov:
Hat tip to @nzakas from eslint/espree who serves as the inspiration for the a bulk of the patch.

### DIFF
--- a/esprima.js
+++ b/esprima.js
@@ -738,6 +738,13 @@
                     extra.openCurlyToken = extra.tokens.length;
                 }
             }
+
+            if (code === 123) {
+                state.curlyStack.push('{');
+            } else if (code === 125) {
+                state.curlyStack.pop();
+            }
+
             return {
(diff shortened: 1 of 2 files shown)


esprima.js
```diff
@@ -737,6 +738,13 @@
                     extra.openCurlyToken = extra.tokens.length;
                 }
             }
+
+            if (code === 123) {
+                state.curlyStack.push('{');
+            } else if (code === 125) {
```
jeffmo: I know the code is inconsistent elsewhere, but can you add a comment to remove any doubt that the intended char here is `}`?

jeffmo:
Just a little nit -- otherwise looks good. Go forth and push, sir!

mikesherov:
Fixed by 5a99bb005322407e54142932fea0c868d70fe743

PR #1107, opened 2015-03-10
Ensure all tests of valid source code compare tokenizer output against parse tokens.

mikesherov:
This is a work in progress. Wanted to just publish some POC code for testing `tokenize()` vs. `parse().tokens`. The tests are failing because it already reveals a bug with tokens in the new `Class` implementation.

Feedback welcome.

### DIFF
--- a/test/runner.js
+++ b/test/runner.js
@@ -201,6 +201,30 @@ function testParse(esprima, code, syntax) {
     }
 }
 
+function testTokenizeVsTokens(esprima, code) {
+    var tree, parseTokens, tokenizeTokens, options;
+
+    options = {
+        comment: true,
+        tolerant: true,
+        loc: true,
+        range: true
+    };
+
+    try {
+        tree = esprima.tokenize(code, options);
+        tokenizeTokens = JSON.stringify(tree, null, 4);
+
+        tree = esprima.parse(code, options);
+        parseTokens = JSON.stringify(tree.tokens, null, 4);
+    } catch (e) {
+        throw new NotMatchingError(code, e.toString());
+    }
+    if (parseTokens !== tokenizeTokens) {
+        throw new NotMatchingError(parseTokens, tokenizeTokens);
+    }
+}
+
 function testTokenize(esprima, code, tokens) {
     'use strict';
     var options, expected, actual, tree;
@@ -379,6 +403,7 @@ if (typeof window === 'undefined') {
                 try {
                     if (testCase.hasOwnProperty('tree')) {
                         testParse(esprima, testCase.case, JSON.parse(testCase.tree));
+                        testTokenizeVsTokens(esprima, testCase.case);
                     } else if (testCase.hasOwnProperty('tokens')) {
                         testTokenize(esprima, testCase.case, JSON.parse(testCase.tokens));
                     } else if (testCase.hasOwnProperty('failure')) {
--- a/test/runner.js
+++ b/test/runner.js
@@ -28,7 +28,7 @@
   THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 */
 
-/*jslint browser:true node:true */
+/*jslint evil:true, browser:true, node:true */
 /*global esprima:true, testFixture:true */
 
 var runTests;
@@ -73,13 +73,14 @@ function sortedObject(o) {
     if (o instanceof RegExp) {
         return o;
     }
-    var keys = Object.keys(o);
-    var result = {
-        range: undefined,
-        loc: undefined
-    };
+    var keys = Object.keys(o),
+        result = {
+            range: undefined,
+            loc: undefined
+        };
+
     keys.forEach(function (key) {
-        if (o.hasOwnProperty(key)){
+        if (o.hasOwnProperty(key)) {
             result[key] = sortedObject(o[key]);
         }
     });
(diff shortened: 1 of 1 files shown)


test/runner.js
```diff
@@ -28,7 +28,7 @@
   THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 */
 
-/*jslint browser:true node:true */
+/*jslint evil:true, browser:true, node:true */
```
ariya: Just like in the main parser, we might as well rid JSLint here.

test/runner.js
```diff
@@ -200,6 +201,30 @@ function testParse(esprima, code, syntax) {
...
+        tree = esprima.tokenize(code, options);
+        tokenizeTokens = JSON.stringify(tree, null, 4);
+
+        tree = esprima.parse(code, options);
+        parseTokens = JSON.stringify(tree.tokens, null, 4);
+    } catch (e) {
+        throw new NotMatchingError(code, e.toString());
+    }
+    if (parseTokens !== tokenizeTokens) {
+        throw new NotMatchingError(parseTokens, tokenizeTokens);
```
mikesherov: need to add a heading here so it's clear what's happening if this error occurs.

ariya:
While we have some logic in the tokenizer to infer some sort of context, do we know for sure that the tokens produced by `tokenize` will match the fidelity of `parse`? There might be two possibilities if it is not the case:

* ES6 syntax requires more context to be inferred in the tokenizer
* It is not possible to do so and we have to live the such differences

mikesherov:
@ariya we should strive to have equivalent output, and so if that isn't possible in `tokenize()`, should we consider correctness and always rely on parse?

ariya:
@mikesherov Depends on the consumer. For correctness, rely only on `parse`. For other usages (best-effort tokenization for syntax highlighting, auto-complete, etc, `tokenize` is still useful.

mikesherov:
There are a few cases where result of a `parse` *could* produce an `Identifier` where it is producing a `Keyword`. Both for class and function declaration. That seems to be the extent of it though, given the results of the work I added.

nzakas:
My two cents: Having different tokens returned from each method is confusing as a consumer. If the only argument for differing results is that it's probably "good enough" for some uses, then I'd rather just have `tokenize()` match what `parse()` does. If there's a use case where having different outputs is necessary, then it could make sense to differ.

ariya:
@nzakas It's not a subjective choice, it's the limitation of not knowing the context. In some cases, it can be fixed by inferring enough context. In other cases, it may be impossible to infer (I don't know yet, haven't looked at it in details).

Example (from @mikesherov's description): 

```js
> require('./esprima').parse('class A { static(){} }', { tokens: true}).tokens
[ { type: 'Keyword', value: 'class' },
  { type: 'Identifier', value: 'A' },
  { type: 'Punctuator', value: '{' },
  { type: 'Keyword', value: 'static' },
  { type: 'Punctuator', value: '(' },
  { type: 'Punctuator', value: ')' },
  { type: 'Punctuator', value: '{' },
  { type: 'Punctuator', value: '}' },
  { type: 'Punctuator', value: '}' } ]

> require('./esprima').tokenize('class A { static(){} }')
[ { type: 'Keyword', value: 'class' },
  { type: 'Identifier', value: 'A' },
  { type: 'Punctuator', value: '{' },
  { type: 'Identifier', value: 'static' },
  { type: 'Punctuator', value: '(' },
  { type: 'Punctuator', value: ')' },
  { type: 'Punctuator', value: '{' },
  { type: 'Punctuator', value: '}' },
  { type: 'Punctuator', value: '}' } ]
```

nzakas:
@ariya I understand how basic tokenization is context-free, I've done a fair amount of work in Espree. :) My point is that there is a way to return the same tokens from `tokenize()` (call `parse()` internally, then return the context-aware tokens generated), so my opinion is that the decision should be use-case driven.

mikesherov:
In my example, not impossible. For non classes, we already return keyword rather than identifier for props that share a name with a keyword.

We should be doing the same for class. That is, dumbing down the parser version to return keyword for consistency.

We could then revisit feasibility of improving both and the tradeoff on speed.

ariya:
@nzakas I don't follow. Are you saying that we should implement `tokenize` by wrapping `parse` and collect the tokens?

ariya:
@mikesherov The context here is strict mode. Class is parsed with implicit strict and therefore `static` is recognized/promoted as a keyword instead of a regular identifier. By that time, inside class body handling, the token `static` has been placed in the token array. To demote `static` back to an identifier, you probably need to walk back the token array to the closest `static`.

Unless there is another way, I would say that this is quite a complicated dance and probably not worth it.

mikesherov:
I will play around with this, because it potentially addresses a fairly large hole in the existing test infrastructure.

There are a few options:

1. Dumb down parser for class.
2. Smarten up tokenizer. 
3. allow the tokensVsTokenize tests to be lenient on identifier/keyword mismatches.
4. Separate this set of tests out as something we run as a sanity check before release but don't fail Jenkins on.


I'm going to pursue this a bit further to see if I can address the concerns.

mikesherov:
@ariya, it also may be as simple as saying `lex()` instead of `scanKeyword()` inside class body parsing. I'll dig in a bit here.

nzakas:
@ariya yup, I'm saying do exactly that if that's what people need from `tokenize()`.

ariya:
@nzakas `tokenize` caters different use-case: syntax highlighting, auto-complete, etc, where the code is not necessary valid or complete or both.

ariya:
@mikesherov Technically speaking, we don't have `scanKeyword()`, only `scanIdentifier()`. Or perhaps I misunderstand what you are referring to?

ariya:
@mikesherov This is rather getting outdated already. Do you want to update it again or shall we close it and give it another try at some point?

mikesherov:
I'll update it.

PR #1110, opened 2015-03-11
Remove the compatibility tests.

### LINKED ISSUE #1109: Remove the compatibility tests
Historically, `test/compat.js` is used to verify that Esprima output matches to that of SM AST using the adopted `test/reflect.js` (from Mozilla) as the baseline. With a growing number of in-house unit tests (almost 900 so far) and considering that ES6 features coming in and its syntax being standardized via ESTree, there is little to no benefit of running these compatibility tests anymore.

ariya:
Closes #1109.


mikesherov:
:+1:

PR #1112, opened 2015-03-11
Distinguish between regex scanning and parsing.

### LINKED ISSUE #1111: Distinguish between scanning regex for parsing vs tokenization
Function `scanRegExp()` can return two completely different objects, depending whether it is invoked from `tokenize` and `parse`. This run-time-dependent return value is confusing, for both human and type inference tool.

It shall be separated into two different functions (scanning vs parsing). The decision to use the right one should happen inside `collectRegex()`.

ariya:
Closes #1111

### DIFF
--- a/esprima.js
+++ b/esprima.js
@@ -1216,10 +1216,10 @@
         };
     }
 
-    function scanRegExp() {
-        scanning = true;
+    function parseRegExpLiteral() {
         var start, body, flags, value;
 
+        scanning = true;
         lookahead = null;
         skipComment();
         start = index;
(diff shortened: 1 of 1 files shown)


ikarienator:
I still think they are both `scan`s. I think we could keep only one of them and let the collect method convert one to another type, that will be cleaner.

ariya:
Good call on the conversion path, that will reduce duplication as well.

ariya:
Closing this now as I got a different (and hopefully) idea.

Issue #1041, opened 2015-02-16
Tolerate unclosed block comment


ariya:
Consider the following test code (parsing in tolerant mode):

```js
require('./esprima').parse('x /* foobar', { tolerant: true })
```

The current master branch will thrown an exception due to the error `Line 1: Unexpected token ILLEGAL`.

If this feature is implemented, it will return an AST because the unclosed block comment is tolerated. The error will be still listed in the `errors` array.

(Migrated from https://code.google.com/p/esprima/issues/detail?id=582, as reported by @mrennie)

PR #1115, opened 2015-03-13
Simplify the logic to handle unclosed block comment.

ariya:
There is no need to check for EOF here because it is implied when
the previous while loop terminates.

Refs #1041


ikarienator:
:+1:

Issue #1027, opened 2015-02-12
Unit Tests: ensure all tests are testing tokens and ranges, and not just nodes


mikesherov:
For example, the harmony branch has tests for es6 that doesn't include token and ranges. That fact has already hidden the fact that both template literals and modules had nasty token bugs.

PR #1116, opened 2015-03-13
Promote strict mode reserved word outside isKeyword().

ariya:
This way, isKeyword() and thereby also scanIdentifer() won't need to know
the current strictness.

Refs #1027


ariya:
@mikesherov I'm not sure how this is going to help your tokens checking directly, but it looks good to have the decoupling anyway.

mikesherov:
:+1:

PR #1119, opened 2015-03-14
Destructuring Binding Pattern Step 1

### LINKED ISSUE #1118: ES6: Destructuring Binding Pattern Step #1
The easiest case for binding pattern is when cover grammar is not needed:

1. function, generator and method params (allow top-level default).
2. standalone var/let/const declaration (require top-level default if complex).
3. catch parameters (disallow top-level default).

Support for binding pattern in for-loops and arrow parameters is pending for further support of cover grammars.

Tracking issue: https://github.com/jquery/esprima/issues/1045

ikarienator:
The easiest case for binding pattern is when cover grammar is not
needed:

1. function, generator and method params (allow top-level default).
2. standalone var/let/const declaration (require top-level default if
complex).
3. catch parameters (disallow top-level default).

Support for binding pattern in for-loops and arrow parameters is pending
for further support of cover grammars.

Closes #1118.


ariya:
LGTM :+1:

mikesherov:
LGTM

PR #1121, opened 2015-03-15
Template Strings

mikesherov:
Closes #1074

Issue #1106, opened 2015-03-10
esprima accepts invalid/incomplete string escape sequences


michaelficarra:
Some invalid string escape sequences that esprima currently parses:

```js
"\x";
"\x0";
"\xx";
"\u";
"\u0";
"\ux";
"\u00";
"\u000";
"\8";
"\9";
```

mikesherov:
@michaelficarra thanks for the report. We'll look into it shortly. Are you interested in submitted a patch for this?

michaelficarra:
I tried working on this, but it appears the `invalid-syntax` tests aren't being run. There's no mention of `invalid-syntax` anywhere in `test`. :confused:

mikesherov:
@michaelficarra all the tests run automatically when you do `npm test`, no?

michaelficarra:
@mikesherov: @ikarienator is working on fixing the issue I mentioned above. He'll open an issue for it shortly.

PR #1124, opened 2015-03-17
Octal literals are not allowed in template strings

mikesherov:
As discussed in #1121, legacy octal literals are strictly forbidden in template strings, as per 11.8.6 and 16.1

### DIFF
--- a/esprima.js
+++ b/esprima.js
@@ -225,6 +225,7 @@
         StrictLHSPostfix: 'Postfix increment/decrement may not have eval or arguments operand in strict mode',
         StrictLHSPrefix: 'Prefix increment/decrement may not have eval or arguments operand in strict mode',
         StrictReservedWord: 'Use of future reserved word in strict mode',
+        TemplateOctalLiteral: 'Octal literals are not allowed in template strings.',
         MissingFromClause: 'Missing from clause',
         NoAsAfterImportNamespace: 'Missing as after import *',
         InvalidModuleSpecifier: 'Invalid module specifier',
@@ -289,6 +290,29 @@
 
     function isOctalDigit(ch) {
         return '01234567'.indexOf(ch) >= 0;
+    }
+
+    function octalToDecimal(ch) {
+        // \0 is not octal escape sequence
+        var octal = (ch !== '0'), code = '01234567'.indexOf(ch);
+
+        if (index < length && isOctalDigit(source[index])) {
+            octal = true;
+            code = code * 8 + '01234567'.indexOf(source[index++]);
+
+            // 3 digits are only allowed when string starts
+            // with 0, 1, 2, 3
+            if ('0123'.indexOf(ch) >= 0 &&
+                    index < length &&
+                    isOctalDigit(source[index])) {
+                code = code * 8 + '01234567'.indexOf(source[index++]);
+            }
+        }
+
+        return {
+            code: code,
+            octal: octal
+        };
     }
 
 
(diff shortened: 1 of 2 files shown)


esprima.js
```diff
@@ -291,6 +292,29 @@
         return '01234567'.indexOf(ch) >= 0;
     }
 
+    function octalToDecimal(ch) {
```
jeffmo: Man I really wish we could just pull in 3rd party dependencies for esprima...

jeffmo:
:+1: ship it!

gibson042:
I like the code consolidation, and bonus removal of unused/unnecessary variables. My sole suggestion is renaming `octalToDecimal` to make clear that it only applies within strings (and templates)—that it tries to parse _LegacyOctalEscapeSequence_ but **not** _LegacyOctalIntegerLiteral_.

mikesherov:
@gibson042 will make that change then ship

esprima.js
```diff
@@ -1274,27 +1281,13 @@
 
                     default:
                         if (isOctalDigit(ch)) {
```
ikarienator: In fact, I think you can simply switch for '0' and check if the next character is decimal or not.

test/harmonytest.js
```diff
@@ -1101,34 +1101,34 @@ var harmonyTestFixture = {
             }
         },
 
-        '`\\u{000042}\\u0042\\x42\\u0\\102\\A`': {
+        '`\\u{000042}\\u0042\\x42\\u0\\A\\0`': {
```
michaelficarra: `` `\u0` `` is not legal. See https://github.com/jquery/esprima/issues/1106.

mikesherov: I'll remove `\u0` which was already there.

esprima.js
mikesherov: This is what you the consolidated code already does. Considering the rarity of this path, I think the consolidation clarity trumps the optimization.

ikarienator: No, not only 1 to 7 throws, all 1 to 9 throws.

https://people.mozilla.org/~jorendorff/es6-draft.html#sec-literals-string-literals

<pre>
<i>NonEscapeCharacter</i> ::
  <i>SourceCharacter</i> but not one of <i>EscapeCharacter</i> or <i>LineTerminator</i>

<i>EscapeCharacter</i> ::
  <i>SingleEscapeCharacter</i>
  <i>DecimalDigit</i>
  <b>x</b>
  <b>u</b>
</pre>

mathiasbynens: https://github.com/whatwg/javascript/issues/17

ikarienator: @mathiasbynens Can you elaborate more? Why is that related to template literals where octals has never been defined?

mathiasbynens: It’s just some background on the history of octals/noctals. For a long time, e.g. `\08` in a string literal was supposed to throw an error (per the spec), but engines in the wild treated it as `\x008`.

Let’s hope implementors won’t make the same mistake when it comes to template literals. https://bugs.ecmascript.org/show_bug.cgi?id=1553 is a better link, actually (removed the old one to remove confusion).

mikesherov: https://bugs.ecmascript.org/show_bug.cgi?id=3477 seems to suggest otherwise, that the behavior is web breaking. Not sure what to do here. @ariya?

I'm matching what I saw when I tested v8, that is:

```
`\08`
```

parses fine.

michaelficarra: My vote is to follow the standard. That way, users know what to expect, and they can account for implementation bugs individually.

mikesherov: @michaelficarra I'm legitimately confused. Does the spec specify this as disallowed inside template strings or is it ambiguous. If you're telling me it's clear, I agree with you. Just not sure what to make of all the info.

michaelficarra: `\08` is disallowed always. Doesn't matter if it's in sloppy mode, in a browser, in a string or regexp or template.

mikesherov: OK, will update then.

ariya:
Closing this as wontfix (see #1172 for details).

PR #1129, opened 2015-03-18
Update LICENSE.BSD

StevenAyr:
Added attribution to bring license in line with BSD 2 clause format.

### DIFF
--- a/LICENSE.BSD
+++ b/LICENSE.BSD
@@ -1,3 +1,5 @@
+Copyright (c) jQuery Foundation, Inc. and Contributors, All Rights Reserved.
+
 Redistribution and use in source and binary forms, with or without
 modification, are permitted provided that the following conditions are met:
 


StevenAyr:
Went through and signed the CLA, please recheck.

ariya:
I think the header of `esprima.js` needs to be adjusted as well? It would be weird if the two are not in sync.

mikesherov:
Yes, there are a couple of BSD headers sprinkled throughout @StevenAyr can you get them all?

StevenAyr:
I'm on it.

StevenAyr:
Went through and updated all the BSD license headers. There are some in-line licenses that I assume apply just to particular contributed code that were not changed, and there are some other licenses in there that I'm going to go back and review, but all BSD licenses should now be up to date.

ariya:
I think some of the third-part code (imported) should be touched (`test/3rdparty` mostly).

ariya:
Also, it probably makes sense to do all the changes in just one commit.

StevenAyr:
That makes sense. I'm running out but will put it all together tomorrow.

PR #1145, opened 2015-03-19
Template Strings: fix double scanning of template literals

### LINKED ISSUE #1143: Template literal parsing error
There appear to be some cases that broke with recent changes.

This fails to parse:
```js
if (condition) {
  (`${foo}`)
}
```

However all of these do parse:

```js
if (condition) {
  `${foo}`
}
```
```js
if (condition)
  `${foo}`
```
```js
if (condition)
  (`${foo}`)
```

Namely there seems to be a problem with the interaction of parens & braces with template literals, probably due to #1104.

cc @mikesherov, and @nzakas (espree is failing in the same way)

mikesherov:
Also breaks scanning of { and } into a separate clause in the switch statement
to avoid having to conditional logic related to the stack on unrelated tokens.
This more closely reflects the situation in the master branch as well.

Fixes #1143

### DIFF
--- a/esprima.js
+++ b/esprima.js
@@ -723,27 +723,41 @@
         case 41:   // ) close bracket
         case 59:   // ; semicolon
         case 44:   // , comma
-        case 123:  // { open curly brace
-        case 125:  // } close curly brace
         case 91:   // [
         case 93:   // ]
         case 58:   // :
         case 63:   // ?
         case 126:  // ~
             ++index;
-            if (extra.tokenize) {
-                if (code === 40) {
-                    extra.openParenToken = extra.tokens.length;
-                } else if (code === 123) {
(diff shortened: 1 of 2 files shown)


mikesherov:
@zpao and @jeffmo ping!

zpao:
This seems like it makes sense and it fixes all the cases we had that were failing so :thumbsup:. I am probably the worst person to actually review the code. I can find bugs and merge our fork (sometimes) but I still don't know how esprima*actually* works.

nzakas:
The changes look good to me. Very clean. :)

jeffmo:
:+1:  looks good.

Sounds like @zpao has run it across our codebase with no problems too, which I've found to be a pretty reassuring smoke test

mikesherov:
OK, landing

PR #1149, opened 2015-03-23
fixes 1000: support ES6 import and export declarations

caridy:
Bulk of the work porting import and export declaration from harmony (from espree actually) into the master branch.

notes:

* ported pieces from https://github.com/jquery/esprima/pull/287
* ported pieces from https://github.com/eslint/espree/pull/43
* `sourceType` is a global setting that can be `module`, which enables import and export declarations. espree is more advanced since it actually turn other features on/off depending on that value, not so sure if esprima wants to have something like that. for now, I make it simple and very restrictive, feel free to refactor it.
* function declarations, and class declarations can have id=null based on some previous discussions to support `export default class {}` and `export default function () {}`.
* the hard part was the tests, having tests with the `tree` case specifying the `sourceType` was easy and simple, but having failures to be validated using `sourceType="module"` was a pain, I ended up creating a new type of case called `module` to support parsing in module mode but validating the failure. maybe someone can refactor that to make it more generic, maybe adding the type into the failure case when possible.


esprima.js
```diff
@@ -4450,6 +4550,258 @@
...
+            // export {default} from 'foo';
+            // export {foo} from 'foo';
+            lex();
+            src = parseModuleSpecifier();
+            consumeSemicolon();
+        } else if (isExportFromIdentifier) {
+            // covering:
+            // export {default}; // missing fromClause
+            throwError(lookahead.value ?
+                    Messages.UnexpectedToken : Messages.MissingFromClause, lookahead.value);
```
mikesherov: indented weirdly.

esprima.js
```diff
@@ -2190,6 +2269,13 @@
         return lookahead.type === Token.Keyword && lookahead.value === keyword;
     }
 
+    // Return true if the next token matches the specified contextual keyword
+    // (where an identifier is sometimes a keyword depending on the context)
+
```
mikesherov: remove newline.

esprima.js
```diff
@@ -160,7 +161,15 @@
         VariableDeclaration: 'VariableDeclaration',
         VariableDeclarator: 'VariableDeclarator',
         WhileStatement: 'WhileStatement',
-        WithStatement: 'WithStatement'
+        WithStatement: 'WithStatement',
+        ExportDefaultDeclaration: 'ExportDefaultDeclaration',
```
mikesherov: alphabetize these please.

esprima.js
```diff
@@ -205,7 +214,12 @@
         DuplicateProtoProperty: 'Duplicate __proto__ fields are not allowed in object literals',
         ConstructorSpecialMethod: 'Class constructor may not be an accessor',
         DuplicateConstructor: 'A class may only have one constructor',
-        StaticPrototype: 'Classes may not have static property named prototype'
+        StaticPrototype: 'Classes may not have static property named prototype',
+        MissingFromClause: 'Missing from clause',
```
mikesherov: are all of these error messages verified to match v8?

esprima.js
```diff
@@ -1898,6 +1912,10 @@
         finishProgram: function (body) {
             this.type = Syntax.Program;
             this.body = body;
+            if (sourceType === 'module') {
+                // very restrictive for now
```
mikesherov: can you please reference the estree PR or section that details this please?

mikesherov:
@caridy looking great! I had a couple of small comments. Also, can you please sign the jQuery CLA? http://contribute.jquery.org/CLA/

esprima.js
nzakas: https://github.com/estree/estree/pull/59

esprima.js
caridy: what do you mean?

mikesherov: All Esprima error messages match those thrown by v8, the JS engine used in chrome and node.

caridy: We don't have import/export declarations in V8 yet, nor in any other engine, we can update those messages in the future.

mikesherov: Ah, for ones where v8 has no message, we try to use the appropriate "unexpected token" message.

@ariya, any thoughts here?

esprima.js
caridy: that's double tab for a multi-line statement ;)

esprima.js
caridy: you can expand the diff, I'm using the existing pattern from similar methods (empty line before and after the method definition...

esprima.js
ariya: Any time we update an error message, it may risk breaking (because some projects might rely on the message). However, moving from the generic "unexpected token" is considered less problematic because it can be perceived as an **upgrade** of the mesage.

esprima.js
mikesherov: Ok.

esprima.js
mikesherov: https://github.com/estree/estree/pull/59#issuecomment-86506047

Please update the PR so it always includes sourceType.

esprima.js
mikesherov: Ok, then seems like we want to convert these all into "unexpected token" messages? Or should we go with these for now, @ariya?

esprima.js
mikesherov: Sorry, was looking on my phone. Missed it.

esprima.js
caridy: I can all the errors statement to use `UnexpectedToken`, or I can change all messages to be "unexpected token" so we keep the semantic in the code.

ariya: I like the idea of keeping the semantic.

esprima.js
```diff
@@ -4450,6 +4550,258 @@
...
+            switch (lookahead.value) {
+                case 'let':
+                case 'const':
+                case 'var':
+                case 'class':
+                case 'function':
+                    declaration = parseStatementListItem();
+                    return node.finishExportNamedDeclaration(declaration, specifiers, null);
+                default:
+                    break;
```
gibson042: This `default` is unnecessary.

esprima.js
mikesherov: @caridy yeah, keep the semantic, just make the message be "unexpected token".

esprima.js
mikesherov: @caridy ping on this comment, because this need to get changed before merge.

ariya:
LGTM :+1:

caridy:
@mikesherov can you merge this? I'm about to go on vacation for 2 months, and I will like to get this clear before I go this weekend.

mikesherov:
@caridy, will do.

ariya:
@caridy Have you signed the CLA? Or perhaps the CLA detection bot did not properly pick that up?

caridy:
@ariya I did by email when the transition happened to jquery foundation, and I did it multiple times thru the bot (with n and ñ) it just keep failing on me. Someone will have to do it manually I guess.

ariya:
@caridy That's what I expect. We'd just ignore the bot for the time being. Thanks!

PR #1150, opened 2015-03-24
Fixes #1148: Destructured argument property shorthand value incorrect

### LINKED ISSUE #1148: Destructured argument property shorthand value incorrect
Given this code:

```js
function x({a}) {
  try {
    var {b} = a;
  }
  catch({stack}) {
  }
}
```

Each of the destructured properties in this example has `shorthand` equal to `false`. I believe they should be `true`.

Tested with current Esprima master.

ikarienator:
Closes #1148.


mikesherov:
LGTM

ariya:
:+1:

PR #1158, opened 2015-03-30
ES6: Add support for generators

### LINKED ISSUE #1033: ES6 feature: Generator: Tracking Issue
## Syntax:

Per Rev 32 (Feb 2, 2015) of the draft spec, the relevant grammar is on [Section 14.4](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-generator-function-definitions) *Generator Function Definitions*:

<pre>
<i>GeneratorMethod</i><sub>[Yield]</sub> :
    <b>*</b> <i>PropertyName</i><sub>[?Yield]</sub> <b>(</b> <i>StrictFormalParameters</i><sub>[Yield,GeneratorParameter]</sub> <b>)</b>  <b>{</b> <i>GeneratorBody</i><sub>[Yield]</sub>  <b>}</b>

<i>GeneratorDeclaration</i><sub>[Yield, Default]</sub> :
    <b>function</b> <b>*</b> <i>BindingIdentifier</i><sub>[?Yield]</sub> <b>(</b> FormalParameters<sub>[Yield, GeneratorParameter]</sub> <b>)</b> <b>{</b> <i>GeneratorBody</i><sub>[Yield]</sub> <b>}</b>
    [+Default] <b>function</b> <b>*</b> <b>(</b> <i>FormalParameters</i><sub>[Yield, GeneratorParameter]</sub> <b>)</b> <b>{</b> <i>GeneratorBody</i><sub>[Yield]</sub> <b>}</b>

<i>GeneratorExpression</i> :
    <b>function</b> <b>*</b> <i>BindingIdentifier</i><sub>[Yield]</sub><sub>opt</sub> <b>(</b> <i>FormalParameters</i><sub>[Yield,GeneratorParameter]</sub> ) <b>{</b> <i>GeneratorBody</i><sub>[Yield]</sub> <b>}</b>

<i>GeneratorBody</i><sub>[Yield]</sub> :
    <i>FunctionBody</i><sub>[?Yield]</sub>

<i>YieldExpression</i><sub>[In]</sub> :
    <b>yield</b>
    <b>yield</b> [no LineTerminator here] [Lexical goal InputElementRegExp] <i>AssignmentExpression</i><sub>[?In, Yield]</sub>
    <b>yield</b> [no LineTerminator here] <b>*</b> [Lexical goal InputElementRegExp] <i>AssignmentExpression</i><sub>[?In, Yield]</sub>

<i>AssignmentExpression</i><sub>[In, Yield]</sub> :
    [+Yield] <i>YieldExpression</i><sub>[?In]</sub>
</pre>

## Spec:

https://people.mozilla.org/~jorendorff/es6-draft.html#sec-generator-function-definitions

Relevant AST interfaces (from https://github.com/estree/estree):

Functions:

```js
extend interface Function {
    defaults: [ Expression ];
    rest: Identifier | null;
    generator: boolean;
}
```

```js
interface ArrowExpression <: Function, Expression {
    type: "ArrowExpression";
    params: [ Pattern ];
    defaults: [ Expression ];
    rest: Identifier | null;
    body: BlockStatement | Expression;
    generator: boolean;
    expression: boolean;
}
```

YieldExpression:

```js
interface YieldExpression <: Expression {
    type: "YieldExpression";
    argument: Expression | null;
}
```
See also:
  * Esprima generator support from harmony branch: https://code.google.com/p/esprima/issues/detail?id=287
  * V8 generator support: https://code.google.com/p/v8/issues/detail?id=2355

## Remaining Tasks:

 - [ ] Review V8 error messages on generator early error
 - [ ] Ensure the proper handling of `yield*` (old issue [617](https://code.google.com/p/esprima/issues/detail?id=617))

dhe128:
This pull request allows esprima to generate ASTs of generators and yield expressions.

One thing I noticed is that yield is reserved in esprima even in non-strict mode. It follows that:
1. There's no need to distinguish BindingIdentifier and BindingIdentifier_Yield
2. There's no need to distinguish x and x_GeneratorParameter, for all productions x. Because except for BindingIdentifier_Yield, _Yield carries enough information so that _GeneratorParameter is not needed.

This pull request relies on yield being reserved in non-strict mode. It tracks state.allowYield but not state.isGenerator. It would be great if you can double check that this is the correct approach.

There's no way to produce ClassExpression_Yield and ClassExpression_GeneratorExpression in the spec. So I ignored it.

resolves #1033


mikesherov:
@dhe128 thanks for contributing! Can you please sign our CLA before we can review this? http://contribute.jquery.org/CLA/

ariya:
@dhe128 Just curious, have signed the CLA?

ariya:
@dhe128 Seems that you have sign the CLA, it's just that the bot doesn't pick it up. We're reviewing this and probably land it pretty soon! /cc @jeffmo

esprima.js
```diff
@@ -2029,6 +2031,14 @@
             this.body = body;
             this.finish();
             return this;
+        },
+
+        finishYieldExpression: function (argument, delegate) {
+            this.type = Syntax.YieldExpression;
+            this.arguments = argument;
```
jeffmo: `argument` (singular vs plural)

esprima.js
```diff
@@ -2029,6 +2031,14 @@
...
+            this.delegate = delegate;
```
jeffmo: Mind sending a PR over to https://github.com/estree/estree/ to include this param in the spec?

esprima.js
```diff
@@ -4369,12 +4478,20 @@
...
                 isStatic = false;
                 computed = match('[');
-                key = parseObjectPropertyKey();
-                if (key.name === 'static' && lookaheadPropertyName()) {
+                if (match('*')) {
+                    lex();
+                } else {
+                    key = parseObjectPropertyKey();
+                }
+                if (key && key.name === 'static' && (lookaheadPropertyName() || match('*'))) {
```
jeffmo: I think you can just move this `if` into the body of the immediately previous `else` and forgo the `key &&` here

esprima.js
```diff
@@ -4549,10 +4666,12 @@
         lookahead = null;
         state = {
             allowIn: true,
+            allowYield: false,
             labelSet: {},
             inFunctionBody: false,
             inIteration: false,
             inSwitch: false,
+            isGenerator: false,
```
jeffmo: This state var doesn't seem to be used

esprima.js
```diff
@@ -4636,10 +4755,12 @@
...
```
jeffmo: same

jeffmo:
Can you also add some tests for `yield` inside an arrow function within a generator?

(`yield` in an arrow function should have the same properties as `yield` in the lexical env immediately outside the arrow)

esprima.js
```diff
@@ -187,6 +188,7 @@
         IllegalContinue: 'Illegal continue statement',
         IllegalBreak: 'Illegal break statement',
         IllegalReturn: 'Illegal return statement',
+        IllegalYield: 'Illegal yield expression',
```
ariya: I don't think V8 has this error message yet. We may stick with the generic "Unexpected token..." until one is available.

jeffmo: For sake of understanding -- what systems might break if we deviated a bit from v8 error messages?

I ask because this seems like a strictly more useful error message for the user + it seems like the idea of guaranteeing that errors match is a bit of a moving target (i.e. don't these things change over time in V8?)

ariya: It's cosmetics and not critical. Once V8 has the error message that we want, we can adopt it. In all cases, the consumer will need to adjust accordingly. It's just easier to transition to a specialized V8-compatible message from a generic "Unexpected token..."  than from an arbitrary one.

ariya:
@dhe128 I can help finishing this, based on the excellent review from @jeffmo, if there is no objection so that we can land it soon.

dhe128:
@ariya, thanks for taking over for me! Sorry I haven't been able to dedicate more time to this.

ariya:
This is now tackled at PR #1178.

PR #1165, opened 2015-04-24
Update attribution in the license.

ariya:
This is to recreate PR #1129 and similar license change as intended by @StevenAyr.


mikesherov:
:+1:

PR #1166, opened 2015-04-28
disable early errors in Shift parser in speed comparison

michaelficarra:


ariya:
I'm undecided about this. Other parsers don't have such a distinction, there are valid different use cases in reporting errors vs skipping them. I think most likely we need two different benchmarks for each case.

mikesherov:
I concur. 2 separate benchs seems to make sense to me.

On Wed, Apr 29, 2015 at 12:15 PM, Ariya Hidayat <notifications@github.com>
wrote:

> I'm undecided about this. Other parsers don't have such a distinction,
> there are valid different use cases in reporting errors vs skipping them. I
> think most likely we need two different benchmarks for each case.
>
> --
> Reply to this email directly or view it on GitHub
> <https://github.com/jquery/esprima/pull/1166#issuecomment-97485079>.
>



-- 
Mike Sherov

michaelficarra:
My argument is that we should use whatever mode gives a result the fastest for a known valid program. So for Acorn, we should be using `parse_dammit`.

ariya:
Close this in favor of Shift in both modes (which I'll implement shortly).

PR #1174, opened 2015-05-02
Link directly to Github issue tracker.

ariya:
With the migration, we might as well link directly. This gives the npmjs page on Esprima to display the actual issues and pull requests.


michaelficarra:
:+1:

hzoo:
Can also update the docs as well - https://github.com/jquery/esprima/blob/gh-pages/doc/index.html

PR #1178, opened 2015-05-13
Generator

ariya:
This supercedes #1158. The original code from @dhe128 has been rebased onto the latest master and I've added a few tweaks based on the review from @jeffmo.

The remaining items:
* `delegate` property to be filed on ESTree
* a new test for `yield` inside an arrow function within a generator


ariya:
Now that I'm thinking about it again, I think it should be `delegating` and not `delegate`.
Thoughts @ikarienator @jeffmo?

jeffmo:
FWIW: I find `delegating` slightly (not to be overstated) confusing because it's not immediately clear what the delegator/delegate is in the the context (and the meaning of the variable gets a little fuzzy).

Going further, `delegates` came to mind -- but I realized it's odd because it's unclear that it's referring to the verb (rather than a list of delegates).

Of the names I can think of, `isDelegator` seems most clear to me. `delegating` seems survivable.

nzakas:
Seems like a good discussion to move to the ESTree repo for resolution :)

esprima.js
```diff
@@ -3616,6 +3692,10 @@
         startToken = lookahead;
         token = lookahead;
 
+        if (matchKeyword('yield')) {
```
ikarienator: Do we allow `yield` to be an identifier?

esprima.js
```diff
@@ -3608,6 +3657,33 @@
...
+
+        expectKeyword('yield');
+
+        if (!hasLineTerminator) {
+            delegate = match('*');
+            if (delegate) {
+                lex();
+            }
+
+            if (!match(';') && !match('}') && lookahead.type !== Token.EOF) {
```
ikarienator: `(yield)` and `[yield]` will fail this.

esprima.js
```diff
@@ -3608,6 +3657,33 @@
...
+        expr = new Node();
+
+        if (!state.allowYield) {
+            tolerateUnexpectedToken(lookahead, Messages.IllegalYield);
+        }
+
+        expectKeyword('yield');
+
+        if (!hasLineTerminator) {
+            delegate = match('*');
```
ikarienator: `yield *` must be followed by an argument.

esprima.js
```diff
@@ -4658,7 +4758,11 @@
             }
         }
 
+        previousAllowYield = state.allowYield;
+        state.allowYield = false;
```
ikarienator: This is not quite right. `yield` expression can be in the initializers of non-generator function parameters. By setting `allowYield` to false, we also allows `yield` as identifiers in generators parameters in non strict mode, which is banned by the spec.

Due to this problem, the `GeneratorParameter` flag cannot be omitted.

esprima.js
```diff
@@ -4696,12 +4804,20 @@
                 token = lookahead;
                 isStatic = false;
                 computed = match('[');
-                key = parseObjectPropertyKey();
-                if (key.name === 'static' && lookaheadPropertyName()) {
-                    token = lookahead;
-                    isStatic = true;
-                    computed = match('[');
+                if (match('*')) {
+                    lex();
```
ikarienator: This fails on `*[computedKey](){}`. The handling of `*` should probably happen before `computed = match('[');`.

esprima.js
ariya: > Do we allow yield to be an identifier?

Probably not. Is there any potential compatibility issue, e.g. old code using `yield` as variable name?

esprima.js
ariya: Good call. Let me file a separate enhancement ticket for that!

esprima.js
jeffmo: The spec outlines the rules for it, but yield is a contextual keyword/identifier. Basically, anywhere inside a generator it's a keyword, and I *think* also if it's in an arrow inside a generator (on my phone now but would have to confirm in the spec).

Also beware of 'yield' context in function params space:
https://twitter.com/bterlson/status/603016789379846144

ariya: > The spec outlines the rules for it, but yield is a contextual keyword/identifier.

OK, let's continue at #1186 so that it is not buried/forgotten easily.

esprima.js
ariya: Should this be covered in ##1186 as well?

esprima.js
ariya: Good catch! I've added a commit that tackles this.

esprima.js
ariya: This should be taken care by `tryParseMethodDefinition`. I agree that it would be nice to centralize the handling`*` but this will do for now.

ariya:
Since there seems to be no objection, I'm going to land this soon.

PR #1181, opened 2015-05-19
update license attribute

pgilad:
specifying the type and URL is deprecated:

https://docs.npmjs.com/files/package.json#license
http://npm1k.org/

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,10 +34,7 @@
     "bugs": {
         "url": "https://github.com/jquery/esprima/issues"
     },
-    "licenses": [{
-        "type": "BSD",
-        "url": "https://github.com/jquery/esprima/raw/master/LICENSE.BSD"
-    }],
+    "license": "BSD-2-Clause",
     "devDependencies": {
         "eslint": "~0.19.0",
         "jscs": "~1.12.0",


ariya:
@mikesherov @kborchers Does this type of change requires CLA?

mikesherov:
Should be fine to merge without.

kborchers:
I disagree with Mike. The goal should be a CLA signature for every commit. For a non-creative commit like this, I think the two options are get a CLA signature or if they don't want to sign, someone can make the change for them and credit the original person in the comment. That second option though doesn't give that person the activity but if they want that, they need to sign the CLA

dmethvin:
Agreed with Kris. If we have to look at a list of exceptions and manually decide whether the changes in the commit reach some bar each time we audit, it makes that work harder. Given the change here, I would just make the change yourself and add "Thanks @pgilad" to the commit if we can't get a signature. 

There *is* a mechanism for making CLA exceptions, see [this list](https://github.com/jquery/jquery-license/tree/master/exceptions),  but they should generally be used only for ancient history where the author can't be reached or the code is no longer in the repo so it's not relevant anymore.

ariya:
@pgilad Are you OK with signing the [online CLA](http://contribute.jquery.org/CLA/)?

pgilad:
@ariya I've already signed jQuery CLA (several times). It seems to not detect me ?

ariya:
@pgilad Thanks, I found your name in the CLA list (seems that the CLA check script missed it).

pgilad:
:+1:

PR #1192, opened 2015-06-17
.npmignore tests. Fixes #1191

### LINKED ISSUE #1191: npmignore test?
esprima's tests weigh about 896KB, and are always bundled in its npm package. There's [an npmignore](https://github.com/jquery/esprima/blob/master/.npmignore) but it only ignores parts of the test directory.

Would you be open to a PR that excludes the entire test directory, so the esprima bundle would be around 90% smaller? This would make a real difference for constantly-rebuilt test runners.

tmcw:


tmcw:
I've signed the CLA, so that test should pass if restarted.

package.json
```diff
@@ -56,7 +56,7 @@
     "scripts": {
         "generate-regex": "node tools/generate-identifier-regex.js",
 
-        "test": "node test/run.js && npm run lint && npm run coverage",
+        "test": "if [ -d test ]; then node test/run.js && npm run lint && npm run coverage; fi",
```
ariya: This will run on Windows, I believe.

tmcw: Do you have any suggestions for how to get this working cross-platform?

ariya: Perhaps have a top-level `unit-tests.js` or something similar? This one will first check if `test/run.js` exists and run it. Otherwise, do nothing.

tmcw:
@ariya done

ariya:
Looks good, can you squash it into one commit please? Thanks.

tmcw:
Cool, done.

ariya:
Thanks! I've landed it with a minor tweak (`unit-tests.js` needs to be in `package.json`). However, seems that the version check needs to be folded there somewhere and also `npat` install is still broken. I'm not an npm expert and hence let's just file it as a known issue for now (#1199).

Issue #1066, opened 2015-02-18
Integrate coveralls.io


ariya:
While our test suite (part of `npm test`) will enforce the 100% coverage threshold, the lack of enough coverage will be only visible from the command-line output, i.e. if one looks at Travis CI report in details.

With Coveralls integration, when a pull request contains one or more commits that regress the coverage, the coverage information will be displayed at the pull request page, thereby giving a direct indication of the issue.

ariya:
When the coverage regresses, coveralls.io also has the full log of the coverage information from the tracked source file (in this case, `esprima.js`). See an example of such a report (this is however for a complete 100% coverage): https://coveralls.io/files/872376647.

PR #1195, opened 2015-06-21
Track code coverage with coveralls.io.

ariya:
Note that this is my first attempt on using Coveralls. It may still not work at all.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,8 @@
         "regenerate": "~0.6.2",
         "unicode-7.0.0": "~0.1.5",
         "json-diff": "~0.3.1",
-        "optimist": "~0.6.0"
+        "optimist": "~0.6.0",
+        "coveralls": "~2.11.2"
     },
     "keywords": [
         "ast",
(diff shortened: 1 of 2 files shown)


ariya:
See https://coveralls.io/r/jquery/esprima for the result of the coverage tracking.

package.json
```diff
@@ -69,6 +70,8 @@
         "check-coverage": "istanbul check-coverage --statement 100 --branch 100 --function 100",
 
         "benchmark": "node test/benchmarks.js",
-        "benchmark-quick": "node test/benchmarks.js quick"
+        "benchmark-quick": "node test/benchmarks.js quick",
+
+        "travis": "npm test && cat ./coverage/lcov.info | coveralls"
```
michaelficarra: I hate unnecessary uses of `cat`.

```sh
npm test && coveralls < ./coverage/lcov.info
```

michaelficarra:
I don't understand what value coveralls adds to a project that enforces 100% coverage. Istanbul is already enforcing 100% coverage during CI. :-1:

package.json
ariya: Good point (I thought you like cats.... never mind). Should be fixed now.

ariya:
> I don't understand what value coveralls adds to a project that enforces 100% coverage.

Informational, see #1066 for more details.

mikesherov:
:+1: Immediate feedback to any contributor on pull requests can't hurt.

ariya:
Expect a future pull request from me with regressed code coverage, just to verify that this is working :-)

Issue #1060, opened 2015-02-17
Patterns in for in / of statements


mikesherov:
kamicane wrote in https://code.google.com/p/esprima/issues/detail?id=558

This is on the harmony branch.

This does not parse for for in / of statements:

    for ([a,b] in x) a;
    for ([a,b] of x) a;
    for ({a,b} in x) a;
    for ({a,b} of x) a;

Same for comprehensions:

    [for ([a,b] of x) a];

Esprima errors out with:

   Invalid left-hand side in for-in

for for in statements, and it parses the ArrayPattern / ObjectPattern as ArrayExpression / ObjectExpression in for of statements, while it should maybe probably be:

    "type": "ForInStatement",
    "left": {
      "type": "ArrayPattern"

This is the tree i tested on:

https://github.com/ariya/esprima/tree/f1b3150aa43205d511e3452b8c1cead96d2925d4

ariya:
We don't support the harmony branch anymore (see #1172). Patterns support for for-in has been covered by #1045 and for-of will be part of #1047.

Issue #1047, opened 2015-02-16
ES6 feature: Iteration using for of


ariya:
### Iteration Statements

<pre>
IterationStatement[Yield, Return] :
  do Statement[?Yield, ?Return] while ( Expression[In, ?Yield] ) ;opt
  while ( Expression[In, ?Yield] ) Statement[?Yield, ?Return]
  for ( [lookahead ∉ {let [}] Expression[?Yield]opt ; Expression[In, ?Yield]opt ; Expression[In, ?Yield]opt ) Statement[?Yield, ?Return]
  for ( var VariableDeclarationList[?Yield] ; Expression[In, ?Yield]opt ; Expression[In, ?Yield]opt ) Statement[?Yield, ?Return]
  for ( LexicalDeclaration[?Yield] Expression[In, ?Yield]opt ; Expression[In, ?Yield]opt ) Statement[?Yield, ?Return]
  for ( [lookahead ∉ {let [}] LeftHandSideExpression[?Yield] in Expression[In, ?Yield] ) Statement[?Yield, ?Return]
  for ( var ForBinding[?Yield] in Expression[In, ?Yield] ) Statement[?Yield, ?Return]
  for ( ForDeclaration[?Yield] in Expression[In, ?Yield] ) Statement[?Yield, ?Return]
  for ( [lookahead ≠ let ] LeftHandSideExpression[?Yield] of AssignmentExpression[In, ?Yield] ) Statement [?Yield, ?Return]
  for ( var ForBinding[?Yield] of AssignmentExpression[In, ?Yield] ) Statement[?Yield, ?Return]
  for ( ForDeclaration[?Yield] of AssignmentExpression[In, ?Yield] ) Statement[?Yield, ?Return]

ForDeclaration[Yield] :
  LetOrConst ForBinding[?Yield]

ForBinding[Yield] :
  BindingIdentifier[?Yield]
  BindingPattern[?Yield]
</pre>

### for-of Statements

```
IterationStatement :
  for ( LeftHandSideExpression in Expression ) Statement
  for ( LeftHandSideExpression of AssignmentExpression ) Statement
```

## Spec:

https://people.mozilla.org/~jorendorff/es6-draft.html#sec-for-in-and-for-of-statements

Relevant AST interfaces (from https://github.com/estree/estree):

```js
interface ForOfStatement <: ForInStatement {
    type: "ForOfStatement";
}
```

## Early Errors:

The following early errors should be detected:

```js
[...]

PR #1196, opened 2015-06-22
Additional for-of tests with array pattern and object pattern.

ariya:
Refs #1060
Refs #1047


ikarienator:
I found out that `for (let {x = 3} of 3);` gives `undefined` in the range of assignment pattern. It may not be related to for-of. `let {x = 3} = x;` also have this problem.

ariya:
@ikarienator Thanks and good catch of that undefined range! I filed it as bug #1201.

michaelficarra:
:+1:

Issue #1169, opened 2015-04-29
ES6: Spread Element (for function argument)


markelog:
## Syntax:

To be written.

## Spec:

https://people.mozilla.org/~jorendorff/es6-draft.html#sec-array-initializer

Relevant AST interfaces (from https://github.com/estree/estree):

```js
interface SpreadElement <: Node {
    type: "SpreadElement";
    argument: Expression;
}

extend interface CallExpression {
    arguments: [ Expression | SpreadElement ];
}
```

## Early Errors:

The following early errors should be detected: (to be written)

## Related Tasks:

- [ ] Error message

michaelficarra:
Spread is not just allowed in function argument position, but also as an array initialiser element.

```js
[...a]
```

markelog:
As i see it, spread operator is not supported for the function calls. Arrays and fe are.

markelog:
So for the expression `a(...b)` correct AST should look like that - 

```js
{
  "type": "Program",
  "body": [
    {
      "type": "ExpressionStatement",
      "expression": {
        "type": "CallExpression",
        "callee": {
          "type": "Identifier",
          "name": "a"
        },
        "arguments": [
          {
            "argument": {
              "name": "b",
              "type": "Identifier",
            },
            "type": "SpreadElement",
          }
        ]
      }
    }
  ]
}
```

@mikesherov, @ariya ?

markelog:
Does cherry-picking <a href="https://code.google.com/p/esprima/issues/detail?id=378#hc2">these</a> commits is sufficient?

/cc @Benvie

markelog:
@mikesherov @ariya could one of you confirm my proposal?

ariya:
@markelog Probably not. Much has changed since and master has diverged much vs the old harmony branch.

ariya:
Apparently spread element for array initializer has been done by @ikarienator as part of #1045. Thus, let's scope this only for function argument.

PR #1200, opened 2015-06-23
ES6: spread element for function argument.

ariya:
Refs #1169


ikarienator:
:+1:. LGTM! I wrote a simple range visualizer to check the range https://gist.github.com/ikarienator/a18b30b7186ac2edd464.

Issue #1175, opened 2015-05-02
Comments are attached to wrong place for objects with more than one keys


jifeon:
I've noticed that if my objects contain more than one field, leadingComments are attached to first property instead of block. Example:
```js
extend(/**@lends o#*/{
    p2: null
});
```
parsed as
```
...
"arguments": [
    {
        "type": "ObjectExpression",
        "properties": [ ... ],
        "leadingComments": [
            {
                "type": "Block",
                "value": "*@lends o#",
                "range": [
                    7,
                    22
                ]
            }
        ]
    }
]
...
```
but
```js
extend(/**@lends o#*/{
    p2: null,
    p3: null
});
```
to
```
"arguments": [
    {
        "type": "ObjectExpression",
        "properties": [
            {
                "type": "Property",
                "key": { ... },
                "computed": false,
                "value": { ... },
                "kind": "init",
                "method": false,
                "shorthand": false,
                "leadingComments": [
                    {
                        "type": "Block",
                        "value": "*@lends o#",
                        "range": [
                            7,
                            22
                        ]
                    }
                ]
            },
            {
                "type": "Property",
                ...
            }
        ]
    }
]
```

mrennie:
Perhaps related to issue https://github.com/jquery/esprima/issues/1071?

jifeon:
Maybe, but it can be two different issues.

jifeon:
Another strange behavior: if I have only one documented variable
```js
/**
 * @type {number}
 */
var a = 5;
```
`leadingComments` are attached to `VariableDeclaration`. If I have a sequence of declarations, `leadingComments` are attached to each `VariableDeclarator`. 

```js
/**
 * @type {number}
 */
var a = 5,
    /**
     * @type {number}
     */
    b = 6;
```
I think comments should be always attached to  `VariableDeclarator`

jifeon:
Github makes mistake, PR merged only in my fork

Issue #1071, opened 2015-02-19
Comments are attached to the wrong node with starting function declaration


mrennie:
Parse the following snippet with the online demo:
```javascript
/***/function o() {/***/function f() {};};
```
For the first FunctionDeclaration the comment is attached to its id node (wrong) and for the second the comment is attached to the FunctionDeclaration node (as expected)

mikesherov:
@mrennie, thanks for contributing! We'll be looking into this issue soon!

mrennie:
Here is another simple example:
```javascript
/**/function f() {var a;}
```

Looks like the attaching code is attaching the comments to the first 'finished' node rather than the node they appear immediately before.

Wouldn't it make more sense to attach comments when you start a node? that way we don;t have to keep a stack of nodes and attaching the comments becomes as simple as assigning arrays to the previous (trailing) and next node we are starting (leading).

ariya:
cc @ikarienator, who also knows a lot about comment attachment.

ikarienator:
Which version are you using?

on master, `esprima.parse('/**/ function a() {}', {attachComment: true})` gives me:
```json
{
    "range": [
        5,
        20
    ],
    "type": "Program",
    "body": [
        {
            "range": [
                5,
                20
            ],
            "type": "FunctionDeclaration",
            "id": {
                "range": [
                    14,
                    15
                ],
                "type": "Identifier",
                "name": "a"
            },
            "params": [],
            "defaults": [],
            "body": {
                "range": [
                    18,
                    20
                ],
                "type": "BlockStatement",
                "body": []
            },
            "rest": null,
            "generator": false,
            "expression": false,
            "leadingComments": [
                {
                    "type": "Block",
                    "value": "",
                    "range": [
                        0,
                        4
                    ]
                }
            ]
        }
    ],
    "comments": [
        {
            "type": "Block",
            "value": "",
            "range": [
                0,
                4
            ]
        }
    ]
}
```

mrennie:
```javascript
/**/ function a() {function o() {}}
```
Run on http://esprima.org/demo/parse.html produces:

```json
{
    "range": [
        5,
        35
    ],
    "type": "Program",
    "body": [
        {
            "range": [
                5,
                35
            ],
            "type": "FunctionDeclaration",
            "id": {
                "range": [
                    14,
                    15
                ],
                "type": "Identifier",
                "name": "a",
                "leadingComments": [
                    {
                        "type": "Block",
                        "value": "",
[...]

PR #1205, opened 2015-06-25
Attaching leading comments fixed

jifeon:
pop call is moved because it affects comment which is before current node and there is redundant iteration with the same last node

Closes gh-1175


jifeon:
I've updated PR because found out one more related issue with nested structures. Also I've added tests and now can assert that this PR also closes #1071

ariya:
LGTM.

PR #1209, opened 2015-06-27
Remove support for component.

### LINKED ISSUE #1207: Remove support for component
According to [component2jspm](https://github.com/componentjs/component2jspm), there are several drawbacks of [component](https://github.com/componentjs/component) such as that there is no more development activity. Its users are migrating to other tool (among others, to jspm).

Removing support for component means that we don't need to have and support `component.json`.

ariya:
Closes #1207


mikesherov:
:+1:



Mike Sherov

PR #1211, opened 2015-06-28
Clean-up the tests infrastructure.

ariya:
Remove tests and parts of tests which are intended for in-browser
testing. Remove manual coverage update tool since now we are using
Coveralls. Simplify the check for cyclomatic complexity.


mikesherov:
:+1:



Mike Sherov

PR #1216, opened 2015-06-29
ES6: meta property `new.target`.

### LINKED ISSUE #1203: ES6 feature: Meta property
## Syntax:

<pre>
NewTarget : new . target
</pre>

## Spec:

https://people.mozilla.org/~jorendorff/es6-draft.html#sec-meta-properties

Relevant AST interfaces (from https://github.com/estree/estree):

```js
    type: "MetaProperty";
    meta: Identifier;
    property: Identifier;
}
```

ariya:
Closes #1203.

### DIFF
--- a/test/fixtures/ES6/meta-property/assign-new-target.tree.json
+++ b/test/fixtures/ES6/meta-property/assign-new-target.tree.json
@@ -1,0 +1,151 @@
+{
+    "range": [
+        0,
+        40
+    ],
+    "loc": {
+        "start": {
+            "line": 1,
+            "column": 0
+        },
+        "end": {
+            "line": 3,
+            "column": 1
+        }
+    },
+    "type": "Program",
+    "body": [
+        {
+            "range": [
+                0,
+                40
+            ],
+            "loc": {
+                "start": {
+                    "line": 1,
+                    "column": 0
+                },
+                "end": {
+                    "line": 3,
+                    "column": 1
+                }
+            },
+            "type": "FunctionDeclaration",
+            "id": {
+                "range": [
+                    9,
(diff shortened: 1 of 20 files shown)


test/fixtures/ES6/meta-property/assign-new-target.tree.json
```diff
@@ -0,0 +1,148 @@
...
+                                            "column": 9
+                                        }
+                                    },
+                                    "type": "Identifier",
+                                    "name": "x"
+                                },
+                                "init": {
+                                    "range": [
+                                        27,
+                                        0
```
ikarienator: The range is incorrect.

ikarienator:
0. I think we should add more tests to make sure the expression precedency is correct.
0. `NewTarget` is not allowed in anywhere out side of a function-like. We should check it.

Otherwise LGTM :+1:.

ariya:
For the operator precedence test, what will be a suitable test case?


 --
Ariya Hidayat, twitter.com/AriyaHidayat

test/fixtures/ES6/meta-property/assign-new-target.tree.json
ariya: Oops, good catch, thank you!

ikarienator:
For example, in `new new.target()()`, the first `new` should be combined with the first pair of parentheses but not the second. The result should be a call expression but not a new expression. This can potentially be interfering with the nature of LL(1) parser. We are parsing this case correctly though. Not a problem.

ariya:
I've added the precedence test. Anything else we shall consider before I'm landing this one?

Issue #1023, opened 2015-02-12
As part of CI, run the test suite of major downstream projects


mikesherov:
Known consumers:

 * ESLint
 * JSCS
 * Istanbul
 * Escope (https://github.com/estools/escope)
 * Escodegen (https://github.com/estools/escodegen)
 * jstransform 
 * 6to5?
 * recast?
 * @ariya @ikarienator @jeffmo others?

ariya:
As long eslint is using espree, it won't be easy (requires some tweaking).

nzakas:
The parser is configurable: http://eslint.org/docs/configuring/#specifying-parser

If that's not enough, let me know what you need.

ariya:
@nzakas If that is enough to run ESLint unit tests, I think it should be sufficient.

nzakas:
It probably depends on how you run the tests, but I think it should work. We set the parser in this file: https://github.com/eslint/eslint/blob/master/conf/eslint.json

mikesherov:
@nzakas only problem is that the suite currently relies on behavior only espree provides.

If we're dedicated to doing this, it would make sense for us to begin work on converging the two projects, at least for the parts that aren't add ons like jsx.

nzakas:
Outside of JSX and `globalReturn`, the rest are ES6 features that Esprima harmony already supports.

mikesherov:
Right, my point being we'd need a way to ignore those tests. Thoughts?

ariya:
@mikesherov It might need a simple pre-processing step, e.g. removing files for unsupported tests.

jeffmo:
Can we add `jstransform` and `6to5` in that list too?
(I suppose we need to just get to writing code and pull requesting here...)

mikesherov:
Exactly. @jeffmo add at to the description at will, and let's turn it into a checklist :-)

fkling:
`recast`, although it is using esprima-fb by default.

mikesherov:
@fkling how can we configure it to use regular esprima?

fkling:
@mikesherov: it's `parse` method accepts an `esprima` [option](https://github.com/benjamn/recast/blob/master/lib/options.js). However, the tests don't seem to be configurable, but I guess one could simply alias `esprima-fb` to `esprima` in the test environment. cc @benjamn who wrote recast.

benjamn:
Happy to make the tests configurable, but not immediately sure what the best way to do that is…

ariya:
[...]

PR #1217, opened 2015-06-30
Downstream project test: include esmangle and JSCS.

ariya:
Refs #1023


mikesherov:
:+1: we might want to make a custom command to pass the local bin directly to JSCS

PR #1218, opened 2015-06-30
Travis CI test: exclude Node.js v0.11.

ariya:
It is considered unstable, superceded by the stable v0.12.
And since the tests only run on v0.10 or later, that's what should be
claimed in package.json.


mikesherov:
:+1:

PR #1219, opened 2015-07-01
Profile analysis using V8 sampling profiler.

### LINKED ISSUE #1157: Should we create a profile command in package.json?
We used `tick` in shift-parser-js to profile JavaScript.

ariya:
Closes #1157


mikesherov:
:+1: perhaps you can show me for to interpret the results of the analysis?

ariya:
@mikesherov For a good start: https://developers.google.com/v8/profiler_example.

PR #1220, opened 2015-07-02
Add extra tests for invalid parameters in arrow function.

### LINKED ISSUE #1010: Parameters in ArrowFunctionExpression are not checked against strict mode in certain circumstances
Parameters in ArrowFunctionExpression are not checked against strict mode restrictions if the strict mode is entered by a use strict directive directly inside the body.

For example:
```es6
eval => { "use strict"; }
```
should be an early error.

ArrowFunctionExpression also does not allow duplicate parameters, which is different from other simple parameter list.

ariya:
Closes #1010.


test/fixtures/ES6/arrow-function/invalid-param-strict-mode.js
```diff
@@ -0,0 +1 @@
+eval => {"use strict"}';
```
mikesherov: Is the stray `'` there on purpose?

ariya: > Is the stray ' there on purpose?

Definitely not. Good catch, will fix it.

ikarienator:
LGTM :+1:!

PR #1221, opened 2015-07-02
Ensure a comma between imported default binding and named imports.

### LINKED ISSUE #1046: 'import foo { bar } from "bar";' should trigger an error
The production `import foo { bar } from "bar";` is an invalid production, but parses as if it were `import foo, { bar } from "bar";`, a valid production. (Note the extra comma after the identifier `foo`.)

Visualization of error:

```
import foo, { bar } from "bar"; // correct
import foo  { bar } from "bar"; // incorrect
          ^ comma must be required here, yet is automatically inserted.
```

Affects harmony branch at top of master.

(Migrated from https://code.google.com/p/esprima/issues/detail?id=610, as filed by @impinball)

ariya:
Fixes #1046

### DIFF
--- a/esprima.js
+++ b/esprima.js
@@ -5199,52 +5199,53 @@
     }
 
     function parseImportDeclaration() {
-        var specifiers, src, node = new Node();
+        var specifiers = [], src, node = new Node();
 
         if (state.inFunctionBody) {
             throwError(Messages.IllegalImportDeclaration);
         }
 
         expectKeyword('import');
-        specifiers = [];
 
         if (lookahead.type === Token.StringLiteral) {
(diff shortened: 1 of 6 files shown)


mikesherov:
Looks like you added 2 throws, but only 1 test which I assume is the cause of the coverage regression. Otherwise, nice work!

ariya:
> Looks like you added 2 throws, but only 1 test which I assume is the cause of the coverage regression.

I intentionally wanted to test Coverall's regression warning but looks like it didn't properly kick in and give the warning :-(

In all cases, I'm adding the extra right away.

PR #1222, opened 2015-07-07
Distinguish `yield` as a keyword vs an identifier.

### LINKED ISSUE #1186: Generator: correctly distinguish yield as identifier vs keyword
### Valid Syntax

Nested yield expression:
```js
function *g() { yield yield }
function *g() { yield *yield }
```

`yield` as the name of a generator, declaration and method:
```js
function *yield() {}
({ *yield() {} })
```

`yield` as the default value:
```js
function *g(x = yield){}
function *g() { (x = yield) => {} }
```

Function expression:
```js
function *g(){ var y = function yield(){}; }
function *g(){ var z = function(yield){}; }
```

Property name:
```js
"use strict"; var { yield: x } = foo;
"use strict"; ({ yield() {} })
function *g() { obj.yield(); }
function *g() { yield obj.yield; }
class A extends B { X(){ super.yield } }
```

Object pattern:
```js
function f({yield: y}){}
function *g({yield: y}){}
```

Arrow expression:
```js
(x) => x + yield;
(z) => { yield + z };
function *g() { (x) => x + yield; }
function *g() { (z) => { yield + z }; }
```

Non-strict mode use of `yield` as an identifier:
```js

([yield] = x)
(x = yield) => {}
(yield) => 42
var { x: yield } = foo
var { yield: x } = foo
try {} catch (yield) {}
function f(yield) {}
function yield(){}
(function(yield) {})
(function yield(){})
({ yield() {} })
let yield = 42
function f(...yield) {}
var yield
```

### Parse Failures

`yield` is not allowed as a binding identifier:

```js
function *g() { try {} catch (yield) {} }
function *g() { let yield }
function *g() { var yield }
function *g() { function *yield(){} }
function *g() { return yield.x; }
function *g(yield){}
function *g(a, b, c, ...yield){}
function *g() { (yield) => 42 }
function *g() { (a, b, c, yield) => 42 }
((function*yield(){})
(function *(yield){})
(function *(x, ...yield){})
export default function *yield() {}
```

Strict mode violation (each line is its own example):

```js

"use strict"; ([yield] = x)
"use strict"; (x = yield) => {}
"use strict"; (yield) => 42
"use strict"; var { x: yield } = foo;
"use strict"; try {} catch (yield) {}
"use strict"; function f(yield) {}
function yield(){ "use strict" }
(function yield(){ "use strict" })
"use strict"; function f() { yield }
"use strict"; let yield = 42
"use strict"; function f(...yield) {}
"use strict"; var yield
"use strict"; function *g() { var y = function yield(){}; }
"use strict"; function *g() { var z = function(yield) {} }
```

Yield expression in a wrong place:
```js
var {x: y = yield 3} = z;
(function() { yield 3; })
function* g() { (x = yield 42) => {} }

```

ariya:
`yield` is a keyword in strict mode, inside a generator, or as part of a
yield expression. `yield` may be used as an identifier for the name of
generators, methods, object properties, and function expressions.

Fixes #1186


mikesherov:
This looks ok to me, however I think @ikarienator is better suited to review.

ariya:
@ikarienator and I had an offline chat this afternoon and we agree that we can land this one.

PR #1224, opened 2015-07-08
Yield expression without an expression should have a null argument.

### LINKED ISSUE #1223: Yield expression without an expression should have a null argument
Test case:

```js
> require('esprima').parse('function *g() { yield }').body[0].body.body[0]
```

Actual output:
```js
{ type: 'ExpressionStatement',
  expression: { type: 'YieldExpression', argument: undefined, delegate: false } }
```

Expected:
```js
{ type: 'ExpressionStatement',
  expression: { type: 'YieldExpression', argument: null, delegate: false } }
```

Per ESTree specification:

```js
interface YieldExpression <: Expression {
    type: "YieldExpression";
    argument: Expression | null;
    delegate: boolean;
}
```

ariya:
Fixes #1223

### DIFF
--- a/esprima.js
+++ b/esprima.js
@@ -3759,6 +3759,7 @@
     function parseYieldExpression() {
         var argument, expr, delegate, previousAllowYield;
 
+        argument = null;
         expr = new Node();
 
         expectKeyword('yield');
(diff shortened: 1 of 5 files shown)


ikarienator:
:+1: LGTM!

PR #1228, opened 2015-07-09
Duplicated proto should be checked only for PropertyDefinition with PropertyName.

### LINKED ISSUE #1225: Duplicated proto should be checked only for PropertyDefinition with PropertyName
Test case:

```js
> require('esprima').parse('x = { __proto__: null, get __proto__(){} }')
```

Actual output:
```
Error: Line 1: Duplicate __proto__ fields are not allowed in object literals
```

Expected: a parse tree since the code is valid. It is a `__proto__` property for the first and a getter function for the second, hence there is no duplicate.

See Annex B.3.1:

> It is a Syntax Error if PropertyNameList of PropertyDefinitionList contains any duplicate entries for `__proto__` and at least two of those entries were obtained from productions of the form _PropertyDefinition : PropertyName : AssignmentExpression_.

Thus, other productions such as _MethodDefinition_ (for methods, setters, getters, generators, ..) should not be checked.

ariya:
Other PropertyDefinition productions, such as MethodDefinition and
IdentifierReference, should be excluded from duplicated `__proto__` checks.

Fixes #1225


ikarienator:
LGTM :+1: !

PR #1230, opened 2015-07-11
Fix ES6 Unicode escape not recognized in identifier name.

### LINKED ISSUE #1229: ES6 Unicode escape sequence is not recognized in identifer name
Test case:

```js
> require('esprima').parse('var \\u{41}')
Error: Line 1: Unexpected token ILLEGAL
````

* Expected: a syntax tree
* Actual: an exception

ariya:
Fixes #1229


mikesherov:
:+1:

PR #1235, opened 2015-07-12
Treat await as a reserved word when parsing as a module.

### LINKED ISSUE #1234: Treat await as a reserved word when parsing a module
See 11.6.2.2  on Future Reserved Words:

> `await` is only treated as a _FutureReservedWord_ when _Module_ is the goal symbol of the syntactic grammar.

ariya:
Fixes #1234


jasonLaster:
Looks good. It still hurts me to see invalid syntax. Do you think an await() invocation example is needed?

ariya:
> Do you think an await() invocation example is needed?

Technically speaking, it's not needed since it also goes to the production of a primary expression.

michaelficarra:
@jasonLaster What invalid syntax?

@ariya I don't see any tests for tolerant mode. Otherwise, LGTM.

ariya:
> I don't see any tests for tolerant mode

Good catch! Not sure why it wasn't there but it's now added.

jasonLaster:
@ariya makes sense.

@michaelficarra I was referring fixtures in `/invalid_expression*.js` like `var await`.

PR #1237, opened 2015-07-14
Add Karma TestRunner

jasonLaster:
Update Unit Tests to use karma test-runner. The change, is largely
because we'd like to be able to run the unit tests in the browser
and it's nice to have a configurable test runner.

There were a couple changes to the unit test build order:

**generateTestCase** was moved to it's own tool. `tool/generate-test-case`

**Fixtures**: Fixtures JSON and JS files are preprocessed and loaded via `<script>`
tags.

The JSON fixtures are available via at `__mocks__` and are processed
with `karma-json-processor`. 

The JS fixtures are available via
`__fixtures` and were processed with `karma-js2js-processor` that I
wrote. It saves the JS in a multiline comment that needs to be unwrapped
after page load.

The `enumerateFixtures` function was updated to account for
the changes.

**Mocha** The custom reporter/runner was replaced with
mocha/chai/sinon/leche, which has
a couple of benefits. At the moment, we're mostly benefiting from the
pretty reporter. Down the road, I anticipate this setup allowing the
tests to be more expressive.

**Next Steps / Questions / Concerns:**
+ There are some failing tests that I'd like some help hunting down. I
believe they're unicode related

+ The tests are organized based on the testX functions (testParse,
testModule, testTree, etc). This made sense at the time, but now i'm
wondering if it'd make more sense to organize the test files in terms of
the fixtures (api, comment, ES6) going forward. If so, we can do this in
a next step.

+ The test run time is significantly slower now. Previously, the tests
ran in 1.3 seconds. Now, the tests run in the same amount of time, but
there's a 15 second load time for the script tags. This could be okay,
as karma has a helpful watch feature, but could also be a workflow pain
point.

+ json diffing - I was surprised to see that there wasn't a good json in browser diff library. I'll investigate and probably choose a decent one for showing diffs.

**run locally** @mikesherov, @ariya I'd appreciate it if you guys pull down the branch and play with it for a bit. I don't want to mess up your flow.


jasonLaster:
UPDATE: chrome has 41 errors, which are all some form of encoding related problems. phantomjs, the tests are failing for some reason, but i'm not sure why.

jasonLaster:
FYI, I dropped the multiline commenting trick for a simple JSON.stringify.

mikesherov:
This is starting to shape up nicely. Thanks @jasonLaster ! @ariya any thoughts so far?

ariya:
Looking really good. Wonderful work @jasonLaster!

A couple of discussion points:

* I suggest separating browser-based tests to its npm script, e.g. it can be executed by `npm run browser-tests`. As such, we may not need to have it run by default. It can be set to run on Travis CI and/or manually.

* What's the plan to run the tests with Node.js?

* Try to include the coverage preprocessor (based on Istanbul) for Karma and see if we can properly get the code coverage report.

* I'm not sure we need to run the test with PhantomJS (especially since the command-line run is best served with Node.js). I think we can aim for Firefox and Chrome for this case (and more browsers, once we get this sorted out and could connect to e.g. Sauce Labs).

* Out of curiousity, is Sinon being used anywhere? Is it a dependent of Leche?

* Some of the failing tests are quite mysterious: "AssertionError: expected undefined to equal" (e.g. for invalid-yield-generator-function-declaration.js). There is nothing weird about the test source itself. This warrants some investigation, perhaps there is unkown escaping issue in js2js?

* Don't worry much about JSON diffing for browser tests. If a test fails on a certain browser, we need to troubleshoot it anyway.  A quick diff between the expected vs baseline will not magically accelerate the debugging, a simple indication that the test fails is often sufficient.

* Minor nitpick: run `npm run jscs` to check for the code style.

test/unit-tests.js
```diff
@@ -105,353 +117,88 @@ function hasAttachedComment(syntax) {
...
-        }
-    }
-
-    if (options.loc) {
-        options.source = syntax.loc.source;
-    }
-
-    syntax = sortedObject(syntax);
-    expected = JSON.stringify(syntax, null, 4);
+function generateTestCase(esprima, testCase) {
```
ariya: I believe this function is not needed anymore.

jasonLaster: killing it

jasonLaster:
Updates:

+ Fixtures are built in two files (fixtures_js, fixtures_json). 
The fixtures are built with a gulp task now. I tried using Karma preprocessors, but they're really limited.

+ Re-Added Node. Mocha has a nice JS Api, so it was fairly easy to run the tests directly. 
Each test now requires its own dependencies, so instead of having globals in `unit-tests.js` there are now modules in `unit-tests/lib/*.js`. 

----

Comments:

+ Deep Diffs - Mocha has a nice diff output in node. I think that that should be sufficient, especially based on what you said above.

+ Code Coverage - How is code coverage currently being done? I think this is something that Karma supports with a framework, but I haven't tried to do it yet.

+ Sinon - Sinon is not used yet. I'll remove it from the package.json.

+ Mysterious encoding fails - yea... I'm pretty stumped here. It should be pretty easy to see how the fixtures are built/read, so i'm curious if anything jumps out at you.

+ JSCS - I'll clean it up tomorrow

package.json
```diff
@@ -1,82 +1,106 @@
...
-        "profile": "node --prof test/profile.js && mv isolate*.log v8.log && node-tick-processor",
-        "benchmark": "node test/benchmarks.js",
-        "benchmark-quick": "node test/benchmarks.js quick",
-
-        "coveralls": "coveralls < ./coverage/lcov.info",
-        "downstream": "node test/downstream.js",
-        "travis": "npm test && npm run coveralls && npm run downstream",
-
-        "generate-regex": "node tools/generate-identifier-regex.js"
+  "name": "esprima",
```
jasonLaster: i'll change this back to 4 spaces

package.json
```diff
@@ -1,82 +1,106 @@
...
+    "node-tick-processor": "~0.0.2",
+    "phantomjs": "~1.9.17",
+    "regenerate": "~0.6.2",
+    "sinon": "~1.15.4",
+    "unicode-7.0.0": "~0.1.5",
+    "vinyl": "^0.5.0",
+    "vinyl-fs": "^1.0.0",
+    "vinyl-source-stream": "^1.1.0",
+    "webpack": "^1.10.1"
+  },
```
jasonLaster: there are probably several libraries we're no longer using, but i added while developing

ariya:
It doesn't seem to work from a clean state (after `git clean -xfd .`). `npm test` will complain

```
[18:20:56] Using gulpfile ~/dev/ariya/esprima/master/gulpfile.js
[18:20:56] Starting 'test'...
[18:20:56] Starting 'jsFixtures'...
[18:20:56] Finished 'jsFixtures' after 14 ms
[18:20:56] Starting 'jsonFixtures'...
[18:20:56] Finished 'jsonFixtures' after 7.46 ms
[18:20:56] Starting 'nodeMocha'...
[18:20:56] 'nodeMocha' errored after 88 ms
[18:20:56] Error: Cannot find module '../../dist/fixtures_js.js'
```

ariya:
Also, can we have the browser test run as part of e.g. `npm run browser-tests`? That is consistent with the script patterns we have in `package.json`.

ariya:
Currently code coverage is done by having Istanbul running the test runner. With Karma, can you try the preprocessor technique I've already mentioned earlier?

Here's also my past blog post that covers (pun intended) the subject: http://ariya.ofilabs.com/2013/12/code-coverage-of-mocha-tests-using-istanbul-and-karma.html.

ariya:
> This patch adds several technologies and opinions to the test architecture:

I have no strong preference of one tool or another for the purpose. However, it is definitely concerning that establishing such a test architecture requires pulling a lot of heavy machinery like this. Particularly, we use only a bit of everything, Mocha's `describe`, Chai `expect().to.equal`, and so on and so forth, and yet all the tools need be present.

Let's think through this and let's ensure that the approach is not overkill.

jasonLaster:
### Updates:
Hey guys, just wanted to share a couple of updates from this week of testing hacks.

This branch currently meets these criteria:

1. working node runner
2. working chrome runner
3. karma based istanbul
4. continuous web test runner (watches for esprima or fixture changes and re-runs)

Drawbacks of this PR's implementation:

1. Karma + Istanbul + Webpack + Source Maps is slow
2. Architecture introduces lots of overhead (gulp, webpack, test modules (`test/unit-tests/*`)) is a lot
3. Tries to do a lot. This is subjective, but I feel like this PR has a lot of half-baked opinions on the reporter, runner, etc.


I think this week was pretty successful. I learned a lot about the current test setup. I experimented with a lot of technologies.

I'm pretty confident that we can make some small solid improvements.
Here's what I think we can do first to add a web runner,
then how we can improve it and the reporter going forward.

Sorry for the many essays I've written over the past week, trying to over communicate as there's a lot here.


### Steps for a good web runner:

1. create `tools/generateTestCases.js`
2. create `tools/generateFixtures.js`
3. update `tests/unit-tests.js` to read fixtures
4. create a simple web test runner


### Things to avoid for the web runner:

1. Don't use gulp as a task-runner (package.json and js files will do)
2. Don't use webpack to build a bundle (it's unecessary overhead)


### Next steps, after a web runner is added:

1. Update Runner / Reporter with Mocha
2. Use Karma to run Istanbul
3. Continuous test mode (watch for changes and re-run tests)


### Nitty Gritty

##### 1. create `tools/generateTestCases.js`

Basically, what's already done in the existing PR.
The primary goal is to make `tests/unit-tests.js` the test runner and nothing else.
Essential for running in the browser.

##### 2. create `tools/generateFixtures.js`

Fixtures are currently built with gulp, but could easily be built in a simple node JS file.

The script will read the JS and JSON fixture files and create two fixture files `tests/dist/js_fixtures.js` `tests/dist/json_fixtures.js`.

NOTE: I'm planning on stubbing (module and window) to export the fixtures
in node and the browser environments. We could use a bundler, but I'd prefer to avoid pre-processing and source maps for the benefit of coverage, run times, and developer workflow.

```js
var js_fixtures = {};
js_fixtures['path/to/fixture'] = "function() {}"

// export js_fixtures in either node or browser environment
// sorry for the hacks (trying to avoid using a bundler that woudl require source maps)
if (!module) module = {};
if (!window) window = {};

module.exports = js_fixtures;
window.js_fixtures = js_fixtures;
```

##### 3. update `tests/unit-tests.js` to read fixtures

There are a couple of things that will change here:
1. how modules are loaded
2. how the script exits

There are a lot of things that will stay the same:
1. The test helpers will be untouched
2. the reporter will be untouched


test loading:
```js
if (typeof require !== "undefined") {
    var esprima = require('.../esprima'),
        js_fixtures = require('dist/js_fixtures.js'),
        json_fixtures = require('dist/json_fixtures.js');
}
```

script exit:
```js
if (typeof process !== "undefined") {
  process.exit(failures.length === 0 ? 0 : 1);
}
```


##### 4. create a simple web test runner

Notes:
1. Why not use karma?

A karma framework plugin would work,
but at that point we're just using karma to add script tags, open a browser, and host a server. Not too much.

2. What about a reporter?

Currently the unit-tests reporter writes failure messages with console.log,
but we could add a minimal reporter that writes failure messages to the page.

3. What will the spec runner look like?
```html
<html>
  <head>
    <script src="../esprima.js"></script>
    <script src="dist/js_fixtures.js"></script>
    <script src="dist/json_fixtures.js"></script>
    <script src="unit-tests.js"></script>
  </head>
  <body>
    <div id="failures"></div>
  </body>
</html>
```

4. Why have a minimal web runner if it can't be programatically launched?

The web runner could be used as a dev tool to make it easy to run tests in the browser and debug esprima.

mikesherov:
I like karma because it does do those minimal things you mentioned. It can also run all the generation commands as a karma watcher task, which allows "save + refresh" debugging....

mikesherov:
IMpressive work here @jasonLaster

jasonLaster:
Thanks @mikesherov. Yea, I like karma as well. A simple karma custom framework, just needs to implement a start function, so that's *not* a deal breaker in terms of a V1 or quick V2.

mikesherov:
So just to converge on minimum solution: let's implement simple web runner with karma launcher and Istanbul preprocessing. cool?

jasonLaster:
That sounds good. Generally, I like to work in small tickets. So, I'll try and get a couple small items done  (generateFixtures, generateTestCases) that should set us up to get the test-runner in. 

Mind if we treat this PR as an investigation and close it when we're happy with the plan.

ariya:
I can say that I'm overall happy with the plan and the progress so far. I have no doubt that this will slowly converge to a solution that satisfies everyone. Thanks for putting the effort to get this to where it is @jasonLaster :+1: 

Another approach for this browser-based test workflow is not to modify `test/unit-tests.js`, i.e. let it be used for our Node.js run for now. Implement something like `test/browser-tests.js` while we continue to iterate. At some point, it should be possible to see the common patterns between the two and merge them. While this requires more long-term effort, it has the immediate benefit of reducing the risk w.r.t the current development process. Does it make sense?

And yes, I agree with landing some items piecemeal. generateTestCase and generateFixture look like a no brainer to me.

jasonLaster:
That makes sense to me. 

Closing as I feel good about the direction.

jasonLaster:
I'll be sharing a PR for the web-runner sometime tomorrow, but here's a sneak peek of the [output](https://www.dropbox.com/s/9jaee0teokphp2i/esprima-tests.png?dl=0).

The beautiful thing for me here is that the runner provides an easy way to see all the tests that run/fail and then run through individual examples.

Issue #1099, opened 2015-03-07
ES6 Main Tracking Issue


mikesherov:
## 2.0

 - [x] arrow functions
 - [x] default function parameters ( #1005)
 - [x] binary/octal integer literals
 - [x] shorthand object property notation
 - [x] method definitions (in object literal)
 - [x] regular expression flags
 - [x] extended unicode escapes in strings, templates, and regexps "\u{123456}"
 - [x] implicit octals

## 2.1 (#1067)

 - [x] block-scoped variable declarations (let/const) (#1065)
 - [x] disallow initialiser in for-in head (let / const / var)
 - [x] rest parameters (#1011)
 - [x] computed property names (#1037)
 - [x] remove duplicate object property (besides __proto__) strict mode restriction (#1037)
 - [x] classes (#1001)

## 2.2 (#1161)
 
 - [x] template strings (#1074)
 - [x] destructuring (#1045)
 - [x] import and export declarations (#1000)
 - [x] super calls and super member accesses (#1147)

## 2.3 (#1189)

 - [x] generator functions + yield (#1033)

## 2.4 (#1202)

 - [x] for-of (#1047)
 - [x] spread arguments (#1169)

## 2.5 (#1246)

 - [x] meta property (#1203)
 - [x] yield: keyword vs identifier (#1186)
 - [x] identifier name (#1232 and #1244)
 - [x] await (#1234)

ariya:
Also, take a look at https://github.com/jquery/esprima/blob/master/ChangeLog. 2.0 already has few more stuff there.

mikesherov:
@ariya, does this look right now?

ariya:
There was also *implicit octal* in 2.0. I personally prefer this term (taken from V8), rather than *noctal* (which seems to be a bit [made-up](https://esdiscuss.org/topic/octal-literals-have-their-uses-you-unix-haters-skip-this-one) and has no precedent anywhere else). /cc @ikarienator

ariya:
2.1 partially triggers an early errors for initializer in for-in (only for `let` and `const`, not yet for `var`).

caridy:
2.2: import and export declarations were implemented here: https://github.com/jquery/esprima/pull/1149

ariya:
Per our discussion this morning, I think we should resolve this task since we reach "feature complete" with respect to ES6 support (and thus let's stop indexing up to 2.5). Further defects and enhancement can be tracked separately.

mikesherov:
Lets close it!

PR #1238, opened 2015-07-16
Use the references and terms from the official ES2015 specification.

ariya:
Refs #1099.


mikesherov:
:+1:

jasonLaster:
Nice, I was wondering where the comments were coming from. 

Might be nice to update to // ES2015 11.3.7.

PR #1240, opened 2015-07-17
Add tools/generate-fixtures.js

jasonLaster:
The tool reads the js and json fixtures and builds two fixture files.

I added two npm packages as devDependencies `lodash` and `glob`. They're
both pretty minor and I don't feel strongly about either.

Here are the two resulting gists:
https://gist.github.com/50f41ee3d919b4d76759
https://gist.github.com/7c16f3f107ab805f0993

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,15 +34,16 @@
     },
     "license": "BSD-2-Clause",
     "devDependencies": {
-        "eslint": "~0.23.0",
-        "jscs": "~1.13.1",
-        "istanbul": "~0.3.16",
+        "coveralls": "~2.11.2",
         "escomplex-js": "1.2.0",
-        "regenerate": "~0.6.2",
-        "unicode-7.0.0": "~0.1.5",
+        "eslint": "~0.23.0",
+        "glob": "^5.0.14",
+        "istanbul": "~0.3.16",
+        "jscs": "~1.13.1",
         "json-diff": "~0.3.1",
         "node-tick-processor": "~0.0.2",
-        "coveralls": "~2.11.2"
+        "regenerate": "~0.6.2",
+        "unicode-7.0.0": "~0.1.5"
     },
     "keywords": [
         "ast",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -34,15 +34,17 @@
...
-        "escomplex-js": "1.2.0",
-        "regenerate": "~0.6.2",
-        "unicode-7.0.0": "~0.1.5",
+        "jscs": "~1.13.1",
         "json-diff": "~0.3.1",
+        "lodash": "^3.10.0",
         "node-tick-processor": "~0.0.2",
-        "coveralls": "~2.11.2"
+        "regenerate": "~0.6.2",
+        "unicode-7.0.0": "~0.1.5"
```
jasonLaster: npm likes to alphabetize the packages, i can re-arrange if need be.

ariya: That should be fine. In fact, we may just adopt it as the standard practice.

ariya:
I assume that `test/dist` is ephemeral? Should this directory be added to `.gitignore`?

jasonLaster:
Yes. I'll add dist to .gitignore.

tools/generate-fixtures.js
```diff
@@ -0,0 +1,70 @@
...
+        var content = "";
+        content += "var fixtures_" + ext + " = {};\n";
+        content += fixtures;
+        content += "module.exports = fixtures_" + ext;
+
+        return content;
+    }
+
+    fixtureFilePaths = glob.sync(path.join(__dirname, '../test/fixtures/**/*.' + ext));
+    fixtures = _.map(fixtureFilePaths, renderFixture).join('\n');
```
ariya: I have no objection on the use of lodash/underscore, but if this is where it is used for now, perhaps we should stick the built-in Array#map?

jasonLaster: Yea, admittedly not the most impressive use of the utility library.

jasonLaster: Will drop _ and re-add if there are real wins.

jasonLaster:
updated.

ariya:
:+1:

ariya:
@jasonLaster I've landed it with ammended commit message to refer the issue. You may want to keep that in mind for your future commits.
Thank you!

jasonLaster:
Thanks. will do @ariya

PR #1243, opened 2015-07-20
Fix #1210 - Unify function expression and declaration functions

### LINKED ISSUE #1210: `parseFunctionExpression` and `parseFunctionDeclaration` have duplicate code starting to be hard to maintain
`parseFunctionExpression` and `parseFunctionDeclaration`  are almost identical. There has been a good reason to do that, but as the implementation become more and more complicated, we should start to abstract them.

jasonLaster:
This is a first pass at #1210. 

There were a couple of ways to approach this ticket: break out parsing into small utilities that (parseFunctionIdentifier, parseFunctionBody, etc..) or use one shared function.

The PROs of small utilities were that the separate behavior for expressions and declarations could remain discrete. The downside, is that there were many local variables: ID, MESSAGE, etc that made it hard to have proper utilities.

Curious to see what you think. I expect several iterations

### DIFF
--- a/esprima.js
+++ b/esprima.js
@@ -1994,20 +1994,8 @@
             return this;
         },
 
-        finishFunctionDeclaration: function (id, params, defaults, body, generator) {
-            this.type = Syntax.FunctionDeclaration;
-            this.id = id;
-            this.params = params;
-            this.defaults = defaults;
-            this.body = body;
-            this.generator = generator;
-            this.expression = false;
-            this.finish();
-            return this;
-        },
-
-        finishFunctionExpression: function (id, params, defaults, body, generator) {
-            this.type = Syntax.FunctionExpression;
+        finishFunctionExpression: function (id, type, params, defaults, body, generator) {
+            this.type = type;
             this.id = id;
             this.params = params;
             this.defaults = defaults;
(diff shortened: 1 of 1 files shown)


esprima.js
```diff
@@ -3184,7 +3172,7 @@
             }
             isAssignmentTarget = isBindingElement = false;
             if (matchKeyword('function')) {
-                return parseFunctionExpression();
+                return parseFunctionExpression(new Node(), true, false);
```
jasonLaster: not a fan of these boolean flags

esprima.js
```diff
@@ -4845,63 +4841,8 @@
...
-                if (isRestrictedWord(token.value)) {
-                    firstRestricted = token;
-                    message = Messages.StrictFunctionName;
-                } else if (isStrictModeReservedWord(token.value)) {
-                    firstRestricted = token;
-                    message = Messages.StrictReservedWord;
-                }
-            }
+        if (isDeclaration) {
+            state.allowYield = !isGenerator;
```
jasonLaster: this diff is hard to follow, but I basically added 4 `isDeclaration` checks in `parseFunctionExpression`

esprima.js
ariya: Maybe turning it into an option object?

ariya:
As a side note, those two were pretty similar before I added the proper support for `yield`. After that, it diverges a bit more, making it hard to completely unify it.
Thoughts @ikarienator?

PR #1252, opened 2015-07-23
Import and export list can be followed by a comma.

### LINKED ISSUE #1250: Import and export list can be followed by a comma
Test case #1:

```js
> require('esprima').parse('import {foo,} from "bar"', { sourceType: 'module' })
Error: Line 1: Unexpected token
```

Test case #2:

```js
require('esprima').parse('export {foo,}', { sourceType: 'module' })
Error: Line 1: Unexpected token }
```

* Expected: valid import and export list
* Actual: Error "Unexpected token"

ariya:
Fixes #1250.

### DIFF
--- a/esprima.js
+++ b/esprima.js
@@ -5110,11 +5110,15 @@
         }
 
         expect('{');
-        if (!match('}')) {
-            do {
-                isExportFromIdentifier = isExportFromIdentifier || matchKeyword('default');
-                specifiers.push(parseExportSpecifier());
-            } while (match(',') && lex());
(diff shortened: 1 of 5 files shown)


mikesherov:
:+1:

jasonLaster:
Small point, but it might be nice to test one/all of these invalid syntaxes:

`import {bar,,baz,} from "foo";`
`import {bar,baz,,} from "foo";`
`import {,bar,baz,} from "foo";`

I don't think they should happen because `parseImport...` should probably catch them, but still might be nice

PR #1253, opened 2015-07-23
Prefix reference comments with ECMA-262 specification

jasonLaster:
Added prefixes to make it more clear what the comments were referencing. The README mentions ECMA-262, but I only saw that recently.

Did i ref the issue correctly?


ariya:
:+1:

PR #1254, opened 2015-07-24
Reformat package.json

jasonLaster:
I apologize in advance for this annoying PR. When `npm install` is run, it reformats the package.json file with this [line](https://github.com/npm/npm/blob/36a412d0eb065cdd1e916f6b001c098ff5c7efc3/lib/install.js#L388). There's been a long [thread](https://github.com/npm/npm/pull/3062) over many years and the isaacs does not want to change this. 

I realize that everything else in the project is formatted with 4 spaces, but it *is* frustrating when the npm tool causes nasty diffs. This is not a big deal, but thought I'd raise it nonetheless. I didn't create a backing issue, because it seemed too small. happy to do that if it'd help or add another issue ref.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,84 +1,79 @@
 {
-    "name": "esprima",
-    "description": "ECMAScript parsing infrastructure for multipurpose analysis",
-    "homepage": "http://esprima.org",
-    "main": "esprima.js",
-    "bin": {
-        "esparse": "./bin/esparse.js",
-        "esvalidate": "./bin/esvalidate.js"
-    },
-    "version": "2.4.0",
-    "files": [
-        "bin",
-        "unit-tests.js",
-        "esprima.js"
-    ],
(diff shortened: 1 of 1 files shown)


ariya:
I think it's fine to follow such a style. Is this behavior introduced in some version of `npm` cause I haven't experienced it before?

jasonLaster:
This seems to be pretty old. You have to do `git install --save-dev karma` to run into it.

mathiasbynens:
@jasonLaster meant `npm install --save […]` or `npm install --save-dev […]`. Just `npm install` doesn’t update `package.json`, indeed.

ariya:
That explains, I never use `--save-dev` for that particular reason.

jasonLaster:
oops, freudian slip

ariya:
:+1:

Issue #1004, opened 2015-02-08
Use test runner to execute the unit tests


mikesherov:
Investigate the use of common test runner (Karma or Venus.js, or other candidates) to run the unit tests.

The benefits:

* there is no need to write a special test runner
* execute the tests on e.g. Sauce Labs
* simplify the test fixture (individual test and expected result)

ariya:
Note that AFAIK, Karma and Venus.js does not support running the tests on Node.js (or these days, io.js). Thus, at least we need to specialize the test runner for stand-alone JS environment vs browser environment.

ariya:
Some historical context:

* https://code.google.com/p/esprima/issues/detail?id=336
* https://code.google.com/p/esprima/issues/detail?id=478
* https://code.google.com/p/esprima/issues/detail?id=606
* https://code.google.com/p/esprima/issues/detail?id=628

nzakas:
Mocha works for both Node.js and browser tests. We use it for both ESLint and Espree.

mikesherov:
@ariya, then I'll take this one and get mocha set up. Expect PR tomorrow.

ariya:
@mikesherov Shouldn't you start with a test runner first (Karma or Venus.js)? Last time I checked, maybe it's changing already these days, the standard `mocha` runner doesn't automatically run the tests in (different) browsers.

mikesherov:
@ariya, there are multiple concerns:

1. Using a test framework so we can unblock #1025 . For this, I'd like to use mocha.

2. Using a test runner so that local development can test in both node and browser environments. This requires karma, which itself requires a test framework, like mocha.

3. Getting that test runner to run both browser and node envs as part of CI. For this, we'd use saucelabs.

I'd like to tackle test framework first, because given that, test runner is essentially trivial.

mikesherov:
Also, I think espree has already solved this problem in a way that makes tests easy to write and contribute: https://github.com/eslint/espree/blob/master/tests/lib/ecma-features.js

Mocha seems like natural starting point with karma then layered on top.

ariya:
I don't really have any preference of a particular test library, mocha or any other.

Could you do me a favor and take a look at `test/parselibs.js`? It may be easier to see where I'm going after you glance at it. It could also show why I believe a test library does not bring much (not that I object or anything) since at most you will end up using the equality check.

[...]

PR #1255, opened 2015-07-27
Add Generate test case

jasonLaster:
Refs #1004.

### DIFF
--- a/test/unit-tests.js
+++ b/test/unit-tests.js
@@ -345,25 +345,6 @@ function testAPI(esprima, code, expected) {
     }
 }
 
-function generateTestCase(esprima, testCase) {
-    var tree, fileName = testCase.key + ".tree.json";
-    try {
-        tree = esprima.parse(testCase.case, {loc: true, range: true});
-        tree = JSON.stringify(tree, null, 4);
-    } catch (e) {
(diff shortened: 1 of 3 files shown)


test/unit-tests.js
```diff
@@ -415,9 +396,6 @@ Object.keys(cases).forEach(function (key) {
                 testError(esprima, testCase.case, JSON.parse(testCase.failure));
             } else if (testCase.hasOwnProperty('result')) {
                 testAPI(esprima, testCase.run, JSON.parse(testCase.result));
-            } else {
-                console.error('Incomplete test case:' + testCase.key + '. Generating test result...');
```
mikesherov: will this now silently fail?

jasonLaster: nice catch. just re-added.

jasonLaster:
Note, I made `generate-testcase` an npm script, but it could just as easily be required in and called directly by unit-tests. 

```js
  var generateTestCase require('utils/generate-testcase');

jasonLaster:
After thinking about it, I'll open this PR after the browser tests land. It'll be easier to see what makes sense after `utils/evaluate-test` is available.

Issue #1159, opened 2015-04-01
add sourceType attribute to Program node for module vs. script


mikesherov:

jasonLaster:
Hey @mikesherov, mind adding some additional context here? Also, I'd be happy to take the ticket off your hands.

michaelficarra:
@jasonLaster https://github.com/estree/estree/blob/master/es6.md#programs

jasonLaster:
thanks

jasonLaster:
@mikesherov @ariya would something like this suffice?

```diff
        finishProgram: function (body) {
             this.type = Syntax.Program;
             this.body = body;
-            if (sourceType === 'module') {
-                // very restrictive for now
-                this.sourceType = sourceType;
-            }
+            this.sourceType = sourceType;
             this.finish();
             return this;
         },
```

PR #1256, opened 2015-07-27
WIP: Add sourceType field for Program node

jasonLaster:
Refs #1159.

@mikesherov does this look right? 

Needs updated unit test fixtures and regression test fixtures.


mikesherov:
Yes, this is it.

esprima.js
```diff
@@ -2104,10 +2104,7 @@
         finishProgram: function (body) {
             this.type = Syntax.Program;
             this.body = body;
-            if (sourceType === 'module') {
-                // very restrictive for now
-                this.sourceType = sourceType;
-            }
+            this.sourceType = sourceType;
```
ariya: Pass the `sourceType` from the function parameter as all the AST node constructor is not supposed to rely much on global(s).

ariya:
> Needs updated unit test fixtures and regression test fixtures.

Right, that's probably the bulk of the work.

esprima.js
jasonLaster: yup, that seems right.

jasonLaster:
@ariya, I rebuilt the fixtures and the  [output](https://gist.github.com/jasonLaster/1aa2d325c294808fbb74) seemed off. Any thoughts?

jasonLaster:
Just pushed the rebuilt unit-test and 3rd party fixtures. Some of the results are a little surprising. Here's a stats view: https://gist.github.com/4ec33f3ae18a59eb3a58

NOTE: viewing the diff takes awhile as `1272` files changed :/

ariya:
There seems to be a linter error. Also, probably worth rebasing it on the latest master.

jasonLaster:
Will do

ariya:
In some cases, it's the ordering. Back then in the days, the location information such `range` or `loc` used to appear at the end. After some optimization, they are the first. The test suite automatically ignores such a different ordering (afterall, it's just JSON). Yet, generating a new fixture means that the baseline will have the updated ordering (location first). It is safe to ignore those cases.

For some other tests cases, e.g. `test/fixtures/ES6/arrow-function/arrow-with-multiple-arg-and-rest.js`, the new fixture/baseline looks radically different, i.e. not just an addition of `"sourceType": "script"` because that test case was parsed with a non-standard parsing option. In this example, I believe the tokens are supposed to be included as well.

The same goes for test cases which need to be parsed with tolerant mode, e.g. `test/fixtures/tolerant-parse/migrated_0052.js`. It should not yield in a failure (`migrated_0052.failure.json`) since the code can be tolerated enough by Esprima, assuming the `tolerant` option is set during parsing.

I don't know the easiest way to fix this. We may introduction `foo.options.json` for every `foo.js` in case the parsing options are custom.

ariya:
@jasonLaster This should be now covered by this pull request: https://github.com/jquery/esprima/pull/1282.

PR #1257, opened 2015-07-28
Run everything.js as part of the tests.

### LINKED ISSUE #1251: Run everything.js as part of the tests
[everything.js](https://www.npmjs.com/package/everything.js) from @michaelficarra  provides a source that contains every ECMA-262 grammatical production. We should run this tests, both for `es2015-script` and `es2015-module`, as part of a new test:

```
npm run grammar-tests
```

This help catching any regression in our ES6 support.

ariya:
Fixes #1251.


mathiasbynens:
:+1: LGTM.

test/grammar-tests.js
```diff
@@ -0,0 +1,37 @@
...
+  (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+  ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+  (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF
+  THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
+
+var esprima = require('../esprima');
+
+function scriptContent(type) {
```
michaelficarra: This is a poorly named function.

ariya: > This is a poorly named function.

Any suggestion?

ikarienator:
:+1:

test/grammar-tests.js
michaelficarra: `readEverythingJsProgram`

test/grammar-tests.js
```diff
@@ -0,0 +1,37 @@
...
+    return require('fs').readFileSync(require.resolve('everything.js/' + type), 'utf-8');
+}
+
+try {
+    esprima.parse(scriptContent('es2015-script'));
+    esprima.parse(scriptContent('es2015-module'), { sourceType: 'module' });
+} catch (e) {
+    console.error('Error:', e.toString());
+    process.exit(1);
+}
```
jasonLaster: It might also be nice to have fixtures for scripts and modules. That way it's both, does it parse and also, does it parse as we'd expect...

michaelficarra: Then the version would have to be locked down. everything.js doesn't guarantee the file won't change between patches. It only guarantees that it will be a valid program of the given type in the given ECMAScript version.

ariya: Agree with @michaelficarra. I think we have enough coverage from the other 1200 test fixtures that this should not be desperately needed. It is more an independent quick verification to prevent regressions.

PR #1263, opened 2015-07-31
ES6: Fix yield operator precedence.

XeCycle:
Let me know if more is needed; I am aware of only code and tests now.


michaelficarra:
LGTM!

ariya:
Excellent fix, thank you! I will land this squashed into one commit.

Issue #1258, opened 2015-07-30
Share code between Node.js test runner and Karma test runner


ariya:
Current the test runner for Node.js-based command-line (`test/unit-tests.js`) and Karma-based browser tests (`test/browser-tests.js`) do not share a lot of code, although some of the functionalities are similar. This issue tracks the effort to unify same bookkeeping activities in both runners.

jasonLaster:
Here's a local branch that's tracking the work
https://github.com/jasonLaster/esprima/tree/unify-tests

ariya:
Anything else to be done regarding this issue?

jasonLaster:
Done

jasonLaster:
It looks like `generateTestCase` was broken when two runners were unified. I'll have a patch soon.

ariya:
I think we can safely resolve this one.

PR #1264, opened 2015-07-31
Cleanup Test Runners

jasonLaster:
Refs #1258
#### Overview
This is a collection of cleanup items that were related
to creating the Browser Test Runner in isolation, and
now unifying the codebase.

#### The test runners share two utilities:

1. createTestCases
2. evaluateTestCase

#### What is different about the two test runners?
+ The browser test runner has a nested mocha reporter
+ The node test runner has a simple reporter, diffed outputs, and can
generate test cases

#### This is a really big patch, what's actually changed?

+ Not much. I recommend looking at the diff with `?w=1` because evaluate-testcase looks like it changed but the UMD wrapper caused indentation
+ `createTestCases` was extracted from `browser-tests` 
+ `enumerateFixtures` was deleted in `unit-tests.js`
+ `evaluateTestCase` was required in `unit-tests.js` and all of the shared code was dropped
+ With all that said, I apologize that this PR is a lot of code to follow.

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -8,6 +8,7 @@ module.exports = function(config) {
       'node_modules/lodash/index.js',
       'test/dist/fixtures_js.js',
       'test/dist/fixtures_json.js',
+      'test/utils/create-testcases.js',
       'test/utils/evaluate-testcase.js',
       'test/browser-tests.js',
     ],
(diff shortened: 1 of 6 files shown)


ariya:
I checked out the branch and played with it. This seems quite straight forward, albeit its rather big change. I think it's good to merge it as is, unless @jasonLaster you want to tweak it a bit more?

jasonLaster:
Just reviewed. I think it's ready to merge. If there are any tweaks or fixes, I'd be happy to do that if they come up.

ariya:
:+1:  Thank you!

PR #1266, opened 2015-08-02
Fixes #1265: Use JSCS preset `crockford`.

### LINKED ISSUE #1265: Use JSCS preset
Rather than supplying our own JSCS configuration, Esprima can use one of the available [JSCS presets](http://jscs.info/overview.html#preset). Based on its [automatic detection](http://ariya.ofilabs.com/2015/06/detecting-and-automatically-fixing-javascript-code-style.html), the closest preset is `crockford`.

ariya:


ariya:
What do you think @mikesherov?

mikesherov:
Sounds good. I don't see though where it is saving the .jscsrc file in this PR?

ariya:
> I don't see though where it is saving the .jscsrc file in this PR?

Not sure I follow. `.jscs.json` is removed in this commit because it's not needed anymore.

PR #1270, opened 2015-08-05
Fix generateTestCase

jasonLaster:
Refs #1258

When the test-runner logic was consolidated, the generateTestCase function was accidentally dropped. This re-adds the function and modularizes `errorToObject` which is shared between generateTestCase and evaluateTestCase.

Note: istanbul is throwing an error, which i still have to investigate.

### DIFF
--- a/test/utils/error-to-object.js
+++ b/test/utils/error-to-object.js
@@ -1,0 +1,46 @@
+/*
+  Copyright (c) jQuery Foundation, Inc. and Contributors, All Rights Reserved.
+
+  Redistribution and use in source and binary forms, with or without
+  modification, are permitted provided that the following conditions are met:
+
+    * Redistributions of source code must retain the above copyright
+      notice, this list of conditions and the following disclaimer.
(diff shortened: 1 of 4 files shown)


ariya:
LGTM. On Travis CI (https://travis-ci.org/jquery/esprima/jobs/74164351), Istanbul seems to be quite happy.

test/utils/error-to-object.js
```diff
@@ -0,0 +1,53 @@
...
+        root.errorToObject = factory();
+    }
+}(this, function() {
+
+  return function (e) {
+      'use strict';
+      var msg = e.toString();
+
+      // Opera 9.64 produces an non-standard string in toString().
+      if (msg.substr(0, 6) !== 'Error:') {
```
mikesherov: Perhaps time to drop Opera 9?

mathiasbynens: +1

jasonLaster: sounds reasonable to me

ariya:
:+1:

PR #1274, opened 2015-08-08
Normalize regex testing when using the new ES6 regex flags.

### LINKED ISSUE #1271: Regex tests are environment dependent
This is about `test/fixtures/expression/primary/literal/regular-expression/migrated_0003.js` (using the new ES6 regex flag `y`):

```js
var x = /[a-z]/y
```

The expected syntax tree, when using the command-line Node.js runner, looks like:

```json
{
    "type": "Program",
    "body": [
        {
            "type": "VariableDeclaration",
            "declarations": [
                {
                    "type": "VariableDeclarator",
                    "id": {
                        "type": "Identifier",
                        "name": "x"
                    },
                    "init": {
                        "type": "Literal",
                        "value": null,
                        "raw": "/[a-z]/y",
                        "regex": {
                            "pattern": "[a-z]",
                            "flags": "y"
                        }
                    }
                }
            ],
            "kind": "var"
        }
    ]
}
```

However, in the environment that supports that flag (e.g. the current Firefox 39), the produced syntax tree will be like this:

```json
{
    "type": "Program",
    "body": [
        {
            "type": "VariableDeclaration",
            "declarations": [
                {
                    "type": "VariableDeclarator",
                    "id": {
                        "type": "Identifier",
                        "name": "x"
                    },
                    "init": {
                        "type": "Literal",
                        "value": "/[a-z]/y",
                        "raw": "/[a-z]/y",
                        "regex": {
                            "pattern": "[a-z]",
                            "flags": "y"
                        }
                    }
                }
            ],
            "kind": "var"
        }
    ]
}
```

Since the environment supports the flag, the `value` is not `null` anymore.

The test suite should normalize this behavior. Otherwise, running the above test on Firefox will result in the test mistakenly marked as a failure.

ariya:
On an environment that already supports the new ES6 regex flags,
the regex value won't be null. This causes the test to be mistakenly
marked as a failure since the baseline has the null value.

To workaround this, ensure that regex value is always null when using
the new ES6 regex flags. Other comparisons will still happen to ensure
that the regex is handled properly.

Fixes #1271.

### DIFF
--- a/test/utils/evaluate-testcase.js
+++ b/test/utils/evaluate-testcase.js
@@ -69,6 +69,13 @@
         keys.forEach(function (key) {
             if (o.hasOwnProperty(key)) {
                 result[key] = sortedObject(o[key]);
+
+                // Nullify regex value for ES6 regex flags
+                if (key === 'value' && o.hasOwnProperty('regex')) {
+                    if (o.regex.flags.match(/[uy]/)) {
+                        result[key] = null;
+                    }
+                }
             }
         });
         return result;


mathiasbynens:
LGTM.

jasonLaster:
LG

test/utils/evaluate-testcase.js
```diff
@@ -69,6 +69,13 @@
         keys.forEach(function (key) {
             if (o.hasOwnProperty(key)) {
                 result[key] = sortedObject(o[key]);
+
+                // Nullify regex value for ES6 regex flags
+                if (key === 'value' && o.hasOwnProperty('regex')) {
+                    if ('uy'.indexOf(o.regex.flags) >= 0) {
```
michaelficarra: 0. This will only work if there's a single flag or if the flags are exactly `uy`, in that order.
0. Why have this line at all? Just omit any regexp values when the node has a `regex` field.

ariya: I'm hesitant to completely ignore the regex object for everything else, since there is a value doing that. I agree on the limited scope of the check, I'll expand it to account for cases whenever `u` or `y` is present.

Issue #1278, opened 2015-08-12
Mocha Browser Runner does not complete


jasonLaster:
The mocha browser runner does not complete.

![](https://www.dropbox.com/s/xkmad55y5h9dzer/Screenshot%202015-08-12%2012.23.51.jpg?dl=1)

It fails on the ES6 Identifier invalid-lone-surrogate test `var source = '\uD800!';` because the source is not URI encodable. 

Why does mocha care that the source is URI encodable? Because, we're including the source in the test name and test names should be greppable in the URL.

ariya:
@jasonLaster Is this now considered as resolved?

jasonLaster:
Yes

PR #1279, opened 2015-08-12
Fix Mocha browser runner

jasonLaster:
Refs #1278

Mocha fails when it encounters test names that are not URI encodable.
We now check to see if the source is encodabable before we add it to the
test name.


ariya:
LGTM!

Issue #1006, opened 2015-02-08
Extend ExpressionStatement to indicate a directive prologue


ariya:
In the current syntax tree, a directive such as `"use strict"` will appear as `ExpressionStatement`. A tool that consumes the tree and needs to be aware of the strict mode will have to perform an extra step to figure out whether such an ExpressionStatement is representing a directive or not.

In this proposal, a new flag `directive` is added to `ExpressionStatement`. The value is set to `true` for a directive prologue.

The current state:

```
> esprima.parse('"use strict"').body[0]
{ type: 'ExpressionStatement',
  expression: 
   { type: 'Literal',
     value: 'use strict',
     raw: '"use strict"' } }
```

If this proposal is implemented, it becomes:

```
> esprima.parse('"use strict"').body[0]
{ type: 'ExpressionStatement',
  expression: 
   { type: 'Literal',
     value: 'use strict',
     raw: '"use strict"' },
  directive: true }
```

Additional resources:

* ECMAScript 5.1 [Section 14.1](http://es5.github.io/#x14.1) on *Directive Prologues and the Use Strict Directive*
* Previous Esprima discussion: https://code.google.com/p/esprima/issues/detail?id=330.
* SpiderMonkey issue: https://bugzilla.mozilla.org/show_bug.cgi?id=791294

ariya:
Based on the previous discussion (#318), seems that a Boolean flag is not enough since there is still additional work to identify the type of the directive.

The example given by @gibson042:

```javascript
strictMode = currentNode.directive === true &&
    // "A Use Strict Directive may not contain an EscapeSequence or LineContinuation"
    currentNode.raw.slice( 1, -1 ) === "use strict";
```

ariya:
Here is a possible alternative.

Use strict directive:

```javascript
> esprima.parse('"use strict"').body[0]
{ type: 'ExpressionStatement',
  expression: 
   { type: 'Literal',
     value: 'use strict',
     raw: '"use strict"' },
  directive: 'use strict' }
```

Not a directive (the literal is in the first statement):

```javascript
> esprima..parse('x; "use strict"').body[1]
{ type: 'ExpressionStatement',
  expression: 
   { type: 'Literal',
     value: 'use strict',
     raw: '"use strict"' },
  directive: null }
```

For the latter, I am ambivalent whether the value should `null` or an empty string.

michaelficarra:
Has to be `null`, otherwise we couldn't distinguish it from the empty directive.

[...]

PR #1281, opened 2015-08-12
Extend ExpressionStatement to indicate a directive prologue

jasonLaster:
Refs #1006 

Started a WIP PR for 'use strict' directives. This is pending estree/estree#6, but in the meantime I have two questions:

1. does the change belong in `finishExpresionStatement` or should we be passing more metadata down like a strict flag?
2. what's the best way to update the tests? will we run into the same problem as before?

### DIFF
--- a/esprima.js
+++ b/esprima.js
@@ -1992,6 +1992,11 @@
         finishExpressionStatement: function (expression) {
             this.type = Syntax.ExpressionStatement;
             this.expression = expression;
+
+            if (expression.value == 'use strict') {
+                this.directive = expression.raw;
+            }
+
             this.finish();
             return this;
         },


ariya:
It's definitely not just inside `finishExpresionStatement`, not every 'use strict' expression magically becomes a directive. Take a look at e.g. `parseScriptBody` where such a directive is detected.

jasonLaster:
@ariya  oy, sloppy. I'll revisit soon.

michaelficarra:
Also, all directives need this property, not just `use strict` directives.

ariya:
Seems that we want to hold this off since the discussion on ESTree is not coming to a consensus.

ariya:
Per the discussion (here and on IRC), let's close this now and reopen again once we get more consensus on ESTree and a refined directive implementation.

PR #1283, opened 2015-08-14
Ensure that strict mode reserved word binding throw an error.

### LINKED ISSUE #1171: Strict mode only reserved words in AssignmentExpressions do not cause parse errors
<B>What steps will reproduce the problem?</B>
Compare Esprima's behaviour when parsing the following two programs. Thanks to @Edgemaster for the examples.
```js
"use strict";
var public = 1;
```
```js
"use strict";
public = 1;
```

<B>What is the expected output?</B>
A parse failure for both programs.

<B>What do you see instead?</B>
The first program results in a parse failure. The second program is parsed as though 'public' were not a ReservedWord in strict mode. Note that this behaviour does not occur if a non-strict mode only reserved word such as enum is substituted.

<B>What version of the product are you using? On what operating system? Which browser?</B>
Master. This bug can be seen on the online parser demo.

ariya:
Fixes #1171

### DIFF
--- a/esprima.js
+++ b/esprima.js
@@ -3867,9 +3867,14 @@
                 tolerateError(Messages.InvalidLHSInAssignment);
             }
 
-            // ECMA-262 11.13.1
-            if (strict && expr.type === Syntax.Identifier && isRestrictedWord(expr.name)) {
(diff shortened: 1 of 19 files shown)


jasonLaster:
looks good.

Issue #1098, opened 2015-03-07
Ensure all early lexical errors are reported during parsing


mikesherov:
Please see https://github.com/shapesecurity/shift-parser-js/blob/es6/test/early-errors.js and specifically the early errors for `let` and `const`.

nowindowsowrking:
@ariya How do you plan to solve this? This is an very easy thing to fix, but many different ways to do it, so I ask for your thoughts.

PR #1285, opened 2015-08-15
Properly allow or prohibit `let` in a binding identifier/pattern.

### LINKED ISSUE #1048: "var let" should be allowed in non-strict mode
**What steps will reproduce the problem?**

```js
esprima.parse("var let")
```

**What is the expected output?** No error

**What do you see instead?** `Unexpected token let`

**What version of the product are you using? On what operating system? Which
browser?** All versions of esprima after https://github.com/ariya/esprima/commit/2b4bdb04ec#diff-7c9336e4324ea31118ce749be055b6a3R311 was committed

**Please provide any additional information below.**

`let` is tricky, it's not a keyword in ES5 or ES6. In strict mode esprima should complain about it, but in non-strict mode `var let` should be fine. The ES6 spec has a few places where `let` is banned even in non-strict mode (see the note on https://people.mozilla.org/~jorendorff/es6-draft.html#sec-keywords ), but this does not include var declarations.

There is currently a test to make sure that "var let" is an error. This test should make sure that it is not an error instead: https://github.com/ariya/esprima/blob/master/test/test.js#L22252

(Migrated from https://code.google.com/p/esprima/issues/detail?id=588, as filed by @gabelevi)

ariya:
Fixes #1048
Refs #1098


ariya:
Unless there is an objection, I will probably land this pretty soon.

michaelficarra:
Let me review first.

ariya:
> Let me review first.

Sure thing.

michaelficarra:
Seems good. I'm going to open an issue for allowing `let` as an identifier in expression position in sloppy mode.

PR #1291, opened 2015-08-20
Move Karma config file to test/.

ariya:
Refs #1004


ariya:
@jasonLaster Seems that it's the natural/suitable place for `karma.conf.js`.

PR #1292, opened 2015-08-20
Workaround read-only Error#column in Safari >= 8.

### LINKED ISSUE #1284: Creating errors breaks in Safari (strict mode)
Creating an `Error` and assigning to the `column` property breaks on the latest mobile and desktop Safari.  Removing `"use strict"` in esprima is a workaround, but seems like not the ideal solution.

Relevant line:
https://github.com/jquery/esprima/blob/master/esprima.js#L2395

Test case:
```
(function () { "use strict"; var e = new Error(); e.column = 1; })();

--> TypeError: Attempted to assign to readonly property.
```

The effect is that `esprima.parse` on invalid code throws the above `TypeError` rather than the expected `SyntaxError` in Safari.

### LINKED ISSUE #1290: Safari: Error#column is read only
This is related to #1284. See also: https://bugs.webkit.org/show_bug.cgi?id=146047

Some unit tests (that check for invalid syntax) fail on Safari. This is because `column` property of an `Error` object is read only.

Run this on Safari (or check this snippet: http://jsbin.com/pexocaruce/edit?js,console):

```js
function createError(msg) {
  var error = new Error(msg);
  error.column = 42;
  console.log('error.column', error.column);
 return error;
}

function throwError(msg) {
  throw createError(msg);
}

try {
  throwError('Hello Universe');
} catch (e) {
  console.log('e.column', e.column);
}
```

It should say that column is 25, despite the fact that its value is set to 42. Chrome and Firefox behave as expected, showing 42.

ariya:
Fixes #1290
Fixes #1284

### DIFF
--- a/esprima.js
+++ b/esprima.js
@@ -2384,12 +2384,30 @@
         extra.errors.push(error);
     }
 
+    function constructError(msg, column) {
+        var error = new Error(msg);
+        try {
+            throw error;
+        } catch (base) {
+            /* istanbul ignore else */
+            if (Object.create && Object.defineProperty) {
+                error = Object.create(base);
+                Object.defineProperty(error, 'column', { value: column });
+            }
+        } finally {
+            return error;
+        }
+    }
+
     function createError(line, pos, description) {
-        var error = new Error('Line ' + line + ': ' + description);
-        error.index = pos;
+        var msg, column, error;
+
+        msg = 'Line ' + line + ': ' + description;
+        column = pos - (scanning ? lineStart : lastLineStart) + 1;
+        error = constructError(msg, column);
         error.lineNumber = line;
-        error.column = pos - (scanning ? lineStart : lastLineStart) + 1;
         error.description = description;
+        error.index = pos;
         return error;
     }
 


ariya:
Hat tip to @gibson042 for the trick to make this work in Safari!

mikesherov:
:+1:

michaelficarra:
We should only throw the error if we need to. This will ruin the actual location information stored in the stack. I imagine the reason Safari made this property non-writable in the first place was so that re-thrown errors would still have position and stack information from the original throw.

ariya:
> We should only throw the error if we need to. This will ruin the actual location information stored in the stack.

I agree, though I think we ought to do this in the forthcoming big refactoring. Right now, there is a big umbrella catch anyway at the top-level `parse` function. All in all, the stack trace doesn't look so bad:

```
> try { esprima.parse('x*') } catch (e) { console.log(e.stack) }
[Log] constructError@http://localhost:8080/esprima.js:2388:30
createError@http://localhost:8080/esprima.js:2407:31
unexpectedTokenError@http://localhost:8080/esprima.js:2481:24
throwUnexpectedToken@http://localhost:8080/esprima.js:2486:35
parsePrimaryExpression@http://localhost:8080/esprima.js:3280:33
inheritCoverGrammar@http://localhost:8080/esprima.js:2662:24
parseLeftHandSideExpressionAllowCall@http://localhost:8080/esprima.js:3387:39
inheritCoverGrammar@http://localhost:8080/esprima.js:2662:24
parsePostfixExpression@http://localhost:8080/esprima.js:3463:35
parseUnaryExpression@http://localhost:8080/esprima.js:3492:42
isolateCoverGrammar@http://localhost:8080/esprima.js:2644:24
parseBinaryExpression@http://localhost:8080/esprima.js:3624:36
inheritCoverGrammar@http://localhost:8080/esprima.js:2662:24
parseConditionalExpression@http://localhost:8080/esprima.js:3669:35
parseAssignmentExpression@http://localhost:8080/esprima.js:3865:42
isolateCoverGrammar@http://localhost:8080/esprima.js:2644:24
parseExpression@http://localhost:8080/esprima.js:3914:35
parseStatement@http://localhost:8080/esprima.js:4715:31
parseStatementListItem@http://localhost:8080/esprima.js:3958:30
parseScriptBody@http://localhost:8080/esprima.js:5407:47
parseProgram@http://localhost:8080/esprima.js:5423:31
parse@http://localhost:8080/esprima.js:5608:35
eval code
eval@[native code]
```

Issue #1294, opened 2015-08-20
Integrate AppVeyor for Windows testing


ariya:
Without Windows CI, it is impossible to know ahead of time if something is going to break on Windows. [AppVeyor](http://www.appveyor.com/) is a popular solution for this since the service is free for FOSS projects.

ariya:
As an illustration, my quick test with AppVeyor shows that the latest master branch fails on Windows:

```
[00:05:39] C:\projects\esprima\test\dist\fixtures_js.js:290
[00:05:39] fixtures_js['ES6\for-of\unexpected-number'] = "for (const of 42);\n";
[00:05:39]             ^^^^^^^^^^^^^
[00:05:39] SyntaxError: Unexpected token ILLEGAL
[00:05:39]     at exports.runInThisContext (vm.js:73:16)
[00:05:39]     at Module._compile (module.js:443:25)
[00:05:39]     at Object.Module._extensions..js (module.js:478:10)
[00:05:39]     at Module.load (module.js:355:32)
[00:05:39]     at Function.Module._load (module.js:310:12)
[00:05:39]     at Module.require (module.js:365:17)
[00:05:39]     at require (module.js:384:17)
[00:05:39]     at cases.(anonymous function).key (C:\projects\esprima\test\utils\create-testcases.js:31:13)
[00:05:39]     at Object.<anonymous> (C:\projects\esprima\test\utils\create-testcases.js:37:2)
[00:05:39]     at Module._compile (module.js:460:26)
```

I think this is because the path separator is backslash (`\`) and therefore it needs to be escaped properly. /cc @jasonLaster

PR #1295, opened 2015-08-21
Fix fixture generation on Windows.

ariya:
Refs #1004


ariya:
I found this while investigating the use of AppVeyor (#1294).

mikesherov:
:+1:

mikesherov:
Good catch!

ariya:
The failing Travis CI test is due to the failure running JSCS as part of downstream test. It's a false alarm.

PR #1300, opened 2015-08-22
Run downstream tests in a temporary directory.

### LINKED ISSUE #1299: Downstream tests should run in a temporary directory
Some third-party projects (#1023) can not complete their unit tests successfully if the test runs within Esprima's own checkout (even indirectly, under a subdirectory). The right solution is to have that project checked out somewhere else.

ariya:
Fixes #1299


ariya:
I'll land this soon so that future pull requests won't be hampered by a failing test.

PR #1302, opened 2015-08-24
Allow binding pattern with no initializer in a for statement.

### LINKED ISSUE #1301: Binding patterns with no initializer should be allowed in a for statement
The following is mistakenly recognized as invalid:

```js
for (var [x, y] in z);
for (var {x, y} in z);
for (var [x, y] of z);
for (var {x, y} of z);
```

Each is valid, due to the following productions (simplified):

```
for ( var ForBinding in Expression ) Statement

for ( var ForBinding of AssignmentExpression ) Statement

ForBinding :
    BindingIdentifier
    BindingPattern

BindingPattern :
    ObjectBindingPattern
    ArrayBindingPattern
```

ariya:
Fixes #1301

### DIFF
--- a/esprima.js
+++ b/esprima.js
@@ -4010,7 +4010,7 @@
         return node.finishIdentifier(token.value);
     }
 
-    function parseVariableDeclaration() {
+    function parseVariableDeclaration(options) {
         var init = null, id, node = new Node(), params = [];
 
         id = parsePattern(params, 'var');
@@ -4023,18 +4023,18 @@
         if (match('=')) {
             lex();
             init = isolateCoverGrammar(parseAssignmentExpression);
-        } else if (id.type !== Syntax.Identifier) {
+        } else if (id.type !== Syntax.Identifier && !options.inFor) {
             expect('=');
         }
 
         return node.finishVariableDeclarator(id, init);
     }
 
-    function parseVariableDeclarationList() {
+    function parseVariableDeclarationList(options) {
         var list = [];
 
         do {
-            list.push(parseVariableDeclaration());
+            list.push(parseVariableDeclaration({ inFor: options.inFor }));
             if (!match(',')) {
                 break;
             }
@@ -4049,7 +4049,7 @@
 
         expectKeyword('var');
 
-        declarations = parseVariableDeclarationList();
+        declarations = parseVariableDeclarationList({ inFor: false });
 
         consumeSemicolon();
 
@@ -4241,21 +4241,24 @@
                 lex();
 
                 state.allowIn = false;
-                init = init.finishVariableDeclaration(parseVariableDeclarationList());
+                declarations = parseVariableDeclarationList({ inFor: true });
                 state.allowIn = previousAllowIn;
 
-                if (init.declarations.length === 1 && matchKeyword('in')) {
+                if (declarations.length === 1 && matchKeyword('in')) {
+                    init = init.finishVariableDeclaration(declarations);
                     lex();
                     left = init;
                     right = parseExpression();
                     init = null;
-                } else if (init.declarations.length === 1 && init.declarations[0].init === null && matchContextualKeyword('of')) {
+                } else if (declarations.length === 1 && declarations[0].init === null && matchContextualKeyword('of')) {
+                    init = init.finishVariableDeclaration(declarations);
                     lex();
                     left = init;
                     right = parseAssignmentExpression();
                     init = null;
                     forIn = false;
                 } else {
+                    init = init.finishVariableDeclaration(declarations);
                     expect(';');
                 }
             } else if (matchKeyword('const') || matchKeyword('let')) {
--- a/test/fixtures/ES6/binding-pattern/array-pattern/var-for-in.js
+++ b/test/fixtures/ES6/binding-pattern/array-pattern/var-for-in.js
@@ -1,0 +1,1 @@
+for (var [x, y] in z);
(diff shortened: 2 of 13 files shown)


esprima.js
```diff
@@ -4023,18 +4023,20 @@
         if (match('=')) {
             lex();
             init = isolateCoverGrammar(parseAssignmentExpression);
-        } else if (id.type !== Syntax.Identifier) {
-            expect('=');
+        } else {
+            if (!(options && options.inFor) && (id.type !== Syntax.Identifier)) {
```
michaelficarra: Unnecessary parens.

michaelficarra:
I'd like to see failure tests for declarations that have initialisers.

esprima.js
```diff
@@ -4023,18 +4023,20 @@
...
         return node.finishVariableDeclarator(id, init);
     }
 
-    function parseVariableDeclarationList() {
+    function parseVariableDeclarationList(options) {
         var list = [];
 
         do {
-            list.push(parseVariableDeclaration());
+            list.push(parseVariableDeclaration(options));
```
michaelficarra: Passing options through blindly bothers me a bit. Now one function's interface is dependent on the other's. This is hostile to change. I'd reconstruct the options object here.

esprima.js
ariya: This is for pure readability. It's challenging to go through that logic without extra help.

ariya:
> I'd like to see failure tests for declarations that have initialisers.

Good catch, I must have messed up my branch. The changes in the test fixture should be bigger than this (including those failing cases).

esprima.js
ariya: Do you mean something like this?
```js
            list.push(parseVariableDeclaration({ inFor: options && options.inFor }));
```

esprima.js
ariya: On that thought, I think I'm just going to unroll the conditionals (easier to follow):
```
            if (id.type !== Syntax.Identifier) {
                if (!(options && options.inFor)) {
                    expect('=');
                }
            }
```

michaelficarra: That's even worse!

esprima.js
michaelficarra: Yes. Then the APIs are independent.

esprima.js
ariya: > That's even worse!

Suggestion(s)?

esprima.js
```diff
@@ -4023,18 +4023,20 @@
...
+            if (!(options && options.inFor) && (id.type !== Syntax.Identifier)) {
+                expect('=');
+            }
         }
 
         return node.finishVariableDeclarator(id, init);
     }
 
-    function parseVariableDeclarationList() {
+    function parseVariableDeclarationList(options) {
```
michaelficarra: These "options" should not be optional. Pass them in explicitly at every call site.

esprima.js
michaelficarra:
```js
} else if (!options.inFor && id.type !== Syntax.Identifier) {
```

esprima.js
```diff
@@ -4023,18 +4023,20 @@
...
         var list = [];
 
         do {
-            list.push(parseVariableDeclaration());
+            list.push(parseVariableDeclaration(options));
             if (!match(',')) {
```
michaelficarra:
```js
if (options.inFor || !match(',')) {
```

esprima.js
ariya: Ah yeah, that must be combined with `options` in every call site.

esprima.js
ariya: I think at this place, there is no sufficient information yet that it will be a for-in.

The check is carried out a bit later:
```js
                if (declarations.length === 1 && declarations[0].init === null && matchKeyword('in')) {
```

ariya:
Whoa, this breaks Esmangle due to its test [test/compare/hoisting-variable-15.js](https://github.com/estools/esmangle/blob/master/test/compare/hoisting-variable-15.js):

```js
// Currently not optimize for-in with initializer
(function () {
    for (var i = 10 in []);
}());
```

michaelficarra:
Yes, it's going to break a lot of things that relied on this feature.

ariya:
We have two possible choices:

(1) Prohibit initializer within a for-in statement for a faithful ES6 compliance. This also means locking down Esmangle to use only Esprima ~2.5.0 if Esmangle still wants to support mangling ES5 code.

(2) Allow initializer in a for-in for now until we have a way to choose between ES5 vs ES6 in some future version. Thus, no change is needed for Esmangle.

Perhaps there is even choice (3) or more.

/cc @Constellation

esprima.js
michaelficarra: There is now that you pass in that information in `options`.

ariya:
> (2) Allow initializer in a for-in for now until we have a way to choose between ES5 vs ES6 in some future version. Thus, no change is needed for Esmangle.

After some thought, I think I prefer to go down this route. I'll tweak the commit and also file a separate issue on invalid initializer in ES6 mode.

mikesherov:
Good catch on the downstream tests :-) i agree that we shouldn't break downstream and instead opt for "es5 mode". however, there are cases where we've paved over the old es5 behavior to make way for es6, are we going to revisit those also, or just cases where we know downstream will break?

michaelficarra:
I don't understand why consumers expecting ES5 parsing are being considered here. They should be using the 1.x branch, right?

ariya:
> however, there are cases where we've paved over the old es5 behavior to make way for es6, are we going to revisit those also, or just cases where we know downstream will break?

I don't know. But for this one, I'll spawn the issue as #1309.

ariya:
> I don't understand why consumers expecting ES5 parsing are being considered here.

Perhaps they should not be. But giving a gentle transition opportunity does not hurt.

michaelficarra:
@ariya It's been pretty gentle, as up to this point you've not implemented any of the ES6 breaking changes.

PR #1303, opened 2015-08-24
Delegate additional stress tests to Codeship.

### LINKED ISSUE #1293: Run additional stress tests separately
For every pull request, the build status (success vs failure) from Travis CI is posted after ~15 minutes. This is because of the following additional tests:
* Submitting code coverage report to Codecov (#1215)
* Running unit tests of downstream projects (#1023)

Those extra tests should be executed using a different CI system, e.g. Circle CI. That way, Travis CI will post the build status almost right away (under 2 minutes) while the extra tests are being executed in parallel.

ariya:
Fixes #1293.


ariya:
Hmm, Codeship still hasn't posted the build status. Let me recreate this pull request to provoke it once more.

PR #1307, opened 2015-08-26
Downstream test: no futher need for globally installed Grunt.

ariya:
This was needed for jsfmt, but it's not necessary anymore:
https://github.com/rdio/jsfmt/pull/163

Refs #1023


ariya:
Almost trivial, so I'll land this.

Issue #1280, opened 2015-08-12
Add a source runner


jasonLaster:
When I want to debug something in Esprima, I use a small page w/ esprima loaded. I believe @mikesherov does something similar.

Would it be nice to improve this process with a shared source runner? I've shared what I use here, but it could obviously be a lot better. If we want to move forward, I can polish and add a `package.json` script for launching the runner.

[source runner](https://gist.github.com/jasonLaster/32ece63e5b08698095e6)

![](https://www.dropbox.com/s/77lo07js8u3ie1a/Screenshot%202015-08-12%2017.44.57.jpg?dl=1)

mikesherov:
Now that we can run the test suite in the browser though I don't need my run script!

ariya:
:+1: 

It's useful for debugging! My main workflow is still using Node.js for troubleshooting (via WebStorm, thanks to its excellent V8 debugger support), but we still need such a runner if there is a bug e.g. only affecting a certain browser.

Right now, the practical way is to use http://esprima.org/demo/parse.html, perhaps with your own version of esprima.js. This runner will bridge the gap.

I think we can name the runner as `test/runner.html` or something along that line.

jasonLaster:
@mikesherov, the test suite runs in ~2 seconds each reload, which can be frustrating if you're just looking to test one scenario.

@ariya `parse` is a good example, but being in `gh-pages` it's tough to test your changes. I like the parse options for comments, tolerate errors, etc and the tree viewer. The ideal solution would be to have one demo tool that is in master branch, but I don't wanto necessarily encur more overhead.

ariya:
I'm fine with such a runner. `test/runner.html` is my proposed name, but feel free to suggest a better one :-)

ariya:
@jasonLaster I think this can be safely marked as resolved?

PR #1310, opened 2015-08-26
Add source runner

jasonLaster:
Refs #1280;

This adds a simple interface for seeing the AST output of esprima. The editor is **code mirror** and the json view is **json editor**.  All of the assets are pulled down via **cdnjs** and the setup is minimal.

Next Steps. I'll add checkboxes for esprima options and a viewer for tokens and raw output. 

Benefit. This makes it easy to test a specific use case while you're debugging locally with chrome devtools. The breakpoint photo below is the main example.

### DIFF
--- a/test/runner.html
+++ b/test/runner.html
@@ -1,0 +1,67 @@
+<html>
+    <head>
+        <script src="../esprima.js"></script>
+        <script src="https://cdnjs.cloudflare.com/ajax/libs/jquery/3.0.0-alpha1/jquery.js"></script>
+
+        <!-- JSON Editor  -->
+        <link href="https://cdnjs.cloudflare.com/ajax/libs/jsoneditor/4.2.1/jsoneditor.css" rel="stylesheet" type="text/css">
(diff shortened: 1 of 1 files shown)


test/runner.html
```diff
@@ -0,0 +1,67 @@
...
+        <script src="https://cdnjs.cloudflare.com/ajax/libs/codemirror/5.6.0/codemirror.js"></script>
+        <script src="https://cdnjs.cloudflare.com/ajax/libs/codemirror/5.6.0/mode/javascript/javascript.js"></script>
+
+        <style>
+          .CodeMirror {
+            border-bottom: 1px solid #ccc;
+          }
+
+          #output-tree {
+            height: calc(100% - 110px);
```
jasonLaster: Responsive design :) Full width, Full Height. Just needs one magic number *110*.

ariya:
Seems good to me. One question: as I try it, the code is automatically parsed for the first time and hence the tree just shows (empty object). Am I missing something?

jasonLaster:
@ariya the code is parsed on ENTER. I'll change that to be every keypress.

ariya:
:+1:

Issue #1308, opened 2015-08-26
Release 2.6


ariya:
- [x] Bump version number
- [x] Update ChangeLog
- [x] Update devDependencies
- [x] Check for performance regression
- [x] Tag the repository
- [x] Publish to npm
- [ ] Update web site

ariya:
Performance difference is negligible (tested with V8 shell 4.3):

2.5.0:
```
    Underscore 1.5.2    42.5 KiB      7.72 ms ± 2.81%
      Backbone 1.1.0    58.7 KiB      6.82 ms ± 1.17%
      MooTools 1.4.5   156.7 KiB     34.00 ms ± 1.50%
        jQuery 1.9.1   262.1 KiB     34.33 ms ± 1.05%
          YUI 3.12.0   330.4 KiB     31.13 ms ± 1.64%
 jQuery.Mobile 1.4.2   442.2 KiB    135.91 ms ± 7.65%
       Angular 1.2.5   701.7 KiB    102.80 ms ± 7.90%
                     ------------------------
                      1994.3 KiB    352.71 ms ± 6.44%
```

2.6.0:
```
    Underscore 1.5.2    42.5 KiB      7.91 ms ± 3.24%
      Backbone 1.1.0    58.7 KiB      6.88 ms ± 1.21%
      MooTools 1.4.5   156.7 KiB     33.81 ms ± 1.79%
        jQuery 1.9.1   262.1 KiB     34.66 ms ± 1.28%
          YUI 3.12.0   330.4 KiB     30.56 ms ± 1.27%
 jQuery.Mobile 1.4.2   442.2 KiB    137.70 ms ± 7.86%
       Angular 1.2.5   701.7 KiB    103.03 ms ± 8.36%
                     ------------------------
                      1994.3 KiB    354.55 ms ± 6.72%
```

PR #1312, opened 2015-08-30
Release 2.6

ariya:


ariya:
For detailed performance numbers, check #1308.

mikesherov:
:+1:

mikesherov:
Btw, for testing downstream projects, we may want to investigate the `dont-break` package.

ariya:
For the badges, I don't have any strong opinion in favor or against it. If those badges seem to be too obnoxious, I'm fine with removing them.

ariya:
> Btw, for testing downstream projects, we may want to investigate the dont-break package.

Good idea. I'll probably file it as an issue at some point but feel free to beat me to it.

jasonLaster:
:+1:

PR #1313, opened 2015-09-03
Fix nonsensical argument passing.

ariya:
This was a mistake in the previous commit 3cf4d0b8, as
parseVariableIdentifier and parseObjectPropertyKey expect a different
number of arguments.

Refs #1098.


ariya:
This is pretty straighforward so I'm landing it soon.

PR #1314, opened 2015-09-04
Unit tests: ensure that all tests are testing tokens.

### LINKED ISSUE #1027: Unit Tests: ensure all tests are testing tokens and ranges, and not just nodes

ariya:
Fixes #1027.


ariya:
The failure from Circle CI is a false alarm. Somehow the downstream project test stopped at JSCS and got stucked there.

ariya:
I added the `tokens` field for all the baselines the same way it was done with the previous `sourceType`. If `patch-tokens.js` is the following:

```js
var fs = require('fs');
var esprima = require('./esprima');

function patch_token(fname) {
    var ast, tokens;
    var source = fs.readFileSync(fname, 'utf-8');
    var outfname = fname.replace('.js', '.tree.json');
    if (fname.indexOf('source.js') > 1) {
        source = eval(source + ';source');
        outfname = fname.replace('.source.js', '.tree.json');
    }
    try {
        var tree = JSON.parse(fs.readFileSync(outfname, 'utf-8'));
        ast = esprima.parse(source, {
            range: true,
            loc: true,
            tokens: true,
            tolerant: true
        });
    } catch (e) {
        return false;
    }

    if (!tree) return false;
    if (tree.tokens) return false;
    if (tree.type && tree.type === 'Program') {
        var store = {};

        for (var i in tree) {
            var p = ['errors', 'range', 'loc'].indexOf(i);
            if (p < 0) continue;
            store[i] = tree[i];
            delete tree[i];

        }

        tree.tokens = ast.tokens;
        for (var i in store) {
            tree[i] = store[i];
        }

        var content = JSON.stringify(tree, null, 4);
        fs.writeFileSync(outfname, content);
        return true;
    }

    return false;
}

if (process.argv.length !== 3) {
    console.error('Usage: patch-tokens.js test-case.js');
    process.exit(-1);
}

var fname = process.argv[2];
if (patch_token(fname))
    console.log('patching', fname);
else
    console.log('    ....', fname);
```

then it was executed as follows:
```
find test/fixtures/ -name '*.js' -execdir node /path/to/patch-tokens.js {} \;
```

mikesherov:
:+1:

PR #1316, opened 2015-09-05
Fix the handling of semicolon after newline in a break statement.

### LINKED ISSUE #1044: Semicolon after newline after break should not cause an error
The following input cannot be parsed:

```js
for (var i=0; i<2; i++) {
	if (i == 1)
		break
		;
	else
		console.log(i);
}
```

The semicolon must be associated with the `break` statement. V8 and the EcmaScript specification both agree on this.

Instead, Esprima parses an empty statement after which the `else` causes a parse error.

Tested with Esprima 1.2.2

(Migrated from https://code.google.com/p/esprima/issues/detail?id=599, as filed by @asgerf).

ariya:
Fixes #1044

### DIFF
--- a/esprima.js
+++ b/esprima.js
@@ -4420,11 +4420,7 @@
             if (!(state.inIteration || state.inSwitch)) {
                 throwError(Messages.IllegalBreak);
             }
-
-            return node.finishBreakStatement(null);
-        }
-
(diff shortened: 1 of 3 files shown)


mikesherov:
:+1:

PR #1326, opened 2015-09-15
Use Sauce Labs to run the test suite.

### LINKED ISSUE #1259: Run extensive browser tests on Sauce Labs
Now that we have a Karma-based test runner for browsers (#1004), we should leverage the "browsers zoo" of Sauce Labs to ensure that Esprima unit tests pass with as many browser as possible.

ariya:
Since the FOSS plan of Sauce Labs only allows 5 concurrent browsers, the
tests are splitted to 3 runs with a different configuration.

Fixes #1259


mikesherov:
:+1:

PR #1327, opened 2015-09-18
In non strict mode, allow `let` as an identifier in expression position.

### LINKED ISSUE #1289: allow `let` as an identifier in expression position in sloppy mode
Fails to parse:

```js
for (let a = let;;);
```

```js
for(let {a: a = let};;);
```

```js
for(let [a = let];;);
```

```js
for(var a in let);
```

```js
for(var a of let);
```

```js
(let[a]=0)
```

```js
let.let=0
```

```js
let
```

ariya:
Fixes #1289


mikesherov:
Looking good. Perhaps add a test for use in implicit strict mode inside modules?

esprima.js
```diff
@@ -3955,6 +3956,12 @@
             }
         }
 
+        if (matchKeyword('let') || matchContextualKeyword('let')) {
```
mikesherov: Which test causes `matchKeyword('let')` to return true here?

mikesherov: Consider adding a function called `matchLet` as this logic is used in several places.

esprima.js
```diff
@@ -4095,6 +4100,56 @@
         return list;
     }
 
+
+    function tokenizerState() {
```
mikesherov: Admitting that I'm still relatively green when it comes to understanding every PR, it seems fishy that this change would be the first to require a double lookahead and restore.

michaelficarra: It's necessary. See the two token lookahead in [section 13.5](http://www.ecma-international.org/ecma-262/6.0/index.html#sec-expression-statement) and the associated NOTE text.

mikesherov: TIL

michaelficarra:
I would leave `let` as a Keyword. It makes it easier to stop accepting it outside of strict mode.

esprima.js
ariya: > Which test causes matchKeyword('let') to return true here?

In a case like this `esprima.parse('let x', { sourceType: 'module'})`. See function `advance()`, it automatically upgrades any strict mode reserved word (including `let`) to Keyword.

ariya: > Consider adding a function called matchLet as this logic is used in several places.

I thought about this, but I think it's only used twice and adding an extra mental note (assuming you're already familiar with `matchKeyword` and `matchContextualKeyword`) does not seem to be worth the hassle.

ariya:
> Perhaps add a test for use in implicit strict mode inside modules?

That should be covered by `test/fixtures/ES6/lexical-declaration/module_let.js`.

mikesherov:
OK, LGTM.

mikesherov:
Might want to address @michaelficarra's concerns though.

ariya:
> I would leave let as a Keyword. It makes it easier to stop accepting it outside of strict mode.

I originally tried this and it was too cumbersome (for some forgotten reasons). Now that I understand the situation a bit better, let me look at this again.

ariya:
BTW that Circle CI failed build is a false alarm (the downstream check fails on JSCS, again).

michaelficarra:
JSCS!!!! :confounded: :weary:

ariya:
I found my previous attempt where `let` was still a keyword. Comparing both versions, I think the keyword should be used for now. I have no strong preference of one or another, but reducing the drastic change for the consumer is always a (minor) bonus.

I'll update this pull request shortly.

ariya:
All tests, including the downstream project tests, now pass with flying colors!

esprima.js
```diff
@@ -2746,7 +2741,12 @@
             return parseArrayPattern(params, kind);
         } else if (match('{')) {
             return parseObjectPattern(params, kind);
+        } else if (matchKeyword('let')) {
+            if (kind === 'const' || kind === 'let') {
```
mikesherov: Could this instead be `!== 'var'` and be folded into previous else if?

mikesherov: Never mind.

PR #1329, opened 2015-09-20
fix #1328, fix comment missing in empty BlockStatement

### LINKED ISSUE #1328: comments missing bug in empty BlockStatement
early time, when es6 multiline feature is not available in node.js,  we always using `module:multiline` to write multiline strings(especially  sql),  it make code easy to read, code like:
```
var multiline = require('multiline');

var sql_xxxx = multiline(function () {
/*
SELECT
    columnes ...
FROM
    some_table
WHERE
    ..........
*/
});
```

but when we write testcase using  istanbul (which is writting in `esprima`, `escodegen`),  the comments in the anonymous function will be ate by esprima.

though esprima later support `attachComment` ( `leadingComments` and `tralingComments`) , but still missing the comments in the empty `BlockStatement`

fishbar:
after esprima apply this patch , `escodegen` also should be patched

### DIFF
--- a/esprima.js
+++ b/esprima.js
@@ -1753,6 +1753,7 @@
 
         processComment: function () {
             var lastChild,
+                innerComments,
                 leadingComments,
                 trailingComments,
                 bottomRight = extra.bottomRightStack,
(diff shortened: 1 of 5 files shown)


ariya:
@fishbar Have you signed the CLA?

fishbar:
@ariya   i'v push a new one to trigger the CLA check, and still failed, see the opened pr

PR #1331, opened 2015-09-21
fix #1328, fix comment missing in empty BlockStatement

### LINKED ISSUE #1328: comments missing bug in empty BlockStatement

fishbar:
after esprima apply this patch , escodegen also should be patched

### DIFF
--- a/esprima.js
+++ b/esprima.js
@@ -1753,6 +1753,7 @@
 
         processComment: function () {
             var lastChild,
+                innerComments,
                 leadingComments,
                 trailingComments,
                 bottomRight = extra.bottomRightStack,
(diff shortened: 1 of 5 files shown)


fishbar:
i'v sign the CLA using the same author info in git commit,  why still failed

ariya:
@fishbar I'll take a look.

kborchers:
The CLA checker is designed to look for 2 names to ensure real names are used. If you click the details link, it states "Suspicious name: fish Suspicious name: fishbar" which means the CLA was signed twice with the name fish and fishbar rather than a family name and a given name.

michaelficarra:
@kborchers It's wrong to put restrictions on a person's name. See *[Falsehoods Programmers Believe About Names](http://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/)*. This restriction violates the [jQuery code of conduct](https://jquery.org/conduct/): be welcoming.

kborchers:
@michaelficarra I honestly agree with you and have made this argument multiple times. I will file an issue to have this check removed. The problem we run into is how do we verify this is a legal name on this legal document without getting a paper signature and photo copy of identification. From what I understand, this requirement was an attempt, and again I agree a bad one, at making sure the name was a legal name which we just may not be able to do and will have to take people's word for it.

PR #1341, opened 2015-10-10
Refactor advanceSlash to rely on a list of simplified tokens.

### LINKED ISSUE #1340: Refactor advanceSlash
To determine whether `/` is the beginning of a regular expression or a division operator, `advanceSlash()` function needs to look back and check some previous tokens. To do this, it relies on the tokens collected during the scanning stage.

Technically speaking, it does not need the full information of each token (location, etc). All it needs is the value of the punctuator and keyword token. It is thus possible to modify `advanceSlash` so that it relies on a different list of such _simplified_ token values.

ariya:
Fixes #1340

### DIFF
--- a/esprima.js
+++ b/esprima.js
@@ -760,14 +760,14 @@
 
         case '(':
             if (extra.tokenize) {
-                extra.openParenToken = extra.tokens.length;
+                extra.openParenToken = extra.tokenValues.length;
             }
             ++index;
             break;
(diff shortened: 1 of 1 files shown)


ariya:
This is rather straightforward and thus I'll land it unless there is an objection.

mikesherov:
:+1:

PR #1342, opened 2015-10-12
Add the support for token delegator.

### LINKED ISSUE #1332: API: Token delegator
A consumer might want to process a token as it is being produced by the tokenizer. This can be achieved by using the [delegation pattern](http://best-practice-software-engineering.ifs.tuwien.ac.at/patterns/delegation.html).

Note: this is related to `tokenize()` and not `parse()`.

ariya:
The third parameter to the `tokenize()` function is a callback function
that is invoked every time the tokenizer encounters a new token. The
return value of this callback function will be appended to the list of
tokens to be returned by `tokenize()`.

An example that returns ['answer', '=', '42']:

  esprima.tokenize('answer = 42', { range: true }, function (token) {
    return token.value;
  });

Fixes #1332

### DIFF
--- a/esprima.js
+++ b/esprima.js
@@ -1643,10 +1643,21 @@
                     flags: token.regex.flags
                 };
             }
-            extra.tokens.push(entry);
             if (extra.tokenValues) {
                 extra.tokenValues.push((entry.type === 'Punctuator' || entry.type === 'Keyword') ? entry.value : null);
             }
+            if (extra.tokenize) {
+                if (!extra.range) {
+                    delete entry.range;
+                }
+                if (!extra.loc) {
+                    delete entry.loc;
+                }
+                if (extra.delegate) {
+                    entry = extra.delegate(entry);
+                }
+            }
+            extra.tokens.push(entry);
         }
 
         return token;
@@ -5519,7 +5530,7 @@
         extra.tokens = tokens;
     }
 
-    function tokenize(code, options) {
+    function tokenize(code, options, delegate) {
         var toString,
             tokens;
 
@@ -5558,6 +5569,8 @@
         extra.tokens = [];
         extra.tokenValues = [];
         extra.tokenize = true;
+        extra.delegate = delegate;
+
         // The following two fields are necessary to compute the Regex tokens.
         extra.openParenToken = -1;
         extra.openCurlyToken = -1;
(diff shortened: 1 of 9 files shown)


ariya:
I finally discovered that it is easy enough to test the delegate: extend the existing test functionalities in `testTokenize`. It is simple and quite effective!

ariya:
Also, the shuffling of properties in the test fixture is a harmless side effect (since we need not call `filterTokenLocation()` anymore after the tokenization is completed).

michaelficarra:
No, this is even worse than I imagined. I strongly urge you to either use another strategy.

* conform to the ES6 iterator interface
* allow ES6 promises to be returned
* pass in another callback to the delegator callback for two-way async communication
* *anything* that is more powerful than a simple synchronous callback

mikesherov:
@michaelficarra perhaps an explanation as to why this is so terrible, and an explanation of how your suggestions are more powerful?

michaelficarra:
This doesn't allow lazily pulling tokens and it requires a synchronous delegator.

Let's say I only need to tokenise the first 10% of the file to find the token I want. I don't want to wait for esprima to tokenise the rest of the file to proceed.

Let's say I need to make an API call to determine how to handle a token. But I'm required to return a value immediately! How would I do this without another callback or a Promise return value?

mikesherov:
> Let's say I only need to tokenise the first 10% of the file to find the token I want. I don't want to wait for esprima to tokenise the rest of the file to proceed.

Entirely possible with current implementation. You could pass a delegator to simply report on the tokens it's encountering. I don't need to wait for all of tokenization to finish.

Same goes for the async case, you could easily do async inside the delegator and ignore the return value of the tokenizer.

This implementation allows you to do whatever you want: provide transformers and rely on output of tokenize, or provide a callback if you need to do anything else special.

I'm not saying your solutions aren't more elegant, but I don't think the current implementation is bad and seems to be the most generic solution.

michaelficarra:
Okay, I see. Yeah, this interface should be sufficient then. Thanks for explaining, @mikesherov.

mikesherov:
:+1:

ariya:
To be fair, the current design doesn't permit the tokenizer to abort. It still needs to process the entire source stream, though the delegator can choose to ignore the rest (if it is only interested in the first 10% tokens). As @michaelficarra mentioned, this is the limitation of such a simple, one-way synchronous callback.

For the record, the iterator-style API is still being planned. This is likely coming after the major surgery for the next 3.0. I'll probably file a separate issue to discuss that.

mikesherov:
Can always `throw` in the delegate ;-). But yeah, by that point, we likely need a new API.

PR #1345, opened 2015-10-17
Token delegate should receive all line and block comments.

### LINKED ISSUE #1344: Token delegate should also receive all comments
With the delegate support (#1332), ideally the delegate should be also invoked for every comment encountered in the source, when the `comment` option is set to true.

Test case:
```js
esprima.tokenize('42 // answer', { comment: true }, function (t) {
    console.log(t.value);
});
```

Current result:
```
42
```

Expected:
```
42
answer
```

ariya:
If `tokenize` is called with a delegate and the tokenization options has
`comment` set to true, then the delegate should also receive every
comment in the source.

Fixes #1344


esprima.js
```diff
@@ -408,6 +408,13 @@
             extra.leadingComments.push(comment);
             extra.trailingComments.push(comment);
         }
+        if (extra.tokenize) {
+            comment.type = comment.type + 'Comment';
```
ariya: Note that this is necessary since we re-use the existing comment collection. For each comment, the type is either `Line` or `Block`. However, we mix the comment with other tokens, e.g. `Numeric`, then `Line` or `Block` does not sufficiently indicating that it is a comment. Trailing it with `Comment` gives us either `LineComment` or `BlockComment`.

mikesherov: Funny, we augment the same way when we mix comments into tokens in JSCS :-)

ariya:
It's rather straighforward so likely I'll land it soon.

mikesherov:
Sorry, forgot to thumb! :+1:

PR #1347, opened 2015-10-19
Add MetaProperty Identifiers

### LINKED ISSUE #1338: MetaProperty implemented incorrectly
According to the spec, MetaProperty is:

```
    type: "MetaProperty";
    meta: Identifier;
    property: Identifier;
}
```

However, Esprima uses strings for both `meta` and `property` rather than `Identifier` objects. (Espree also does this incorrectly.)

See http://esprima.org/demo/parse.html?code=function%20f()%20%7B%0A%20%20%20%20var%20x%20%3D%20new.target%3B%0A%7D%0A for an example.

jasonLaster:
resolves #1338 

NOTE: this is a breaking change, so it'll have to wait for 3.0

I'll share a [branch](https://github.com/jquery/esprima/commit/e9b19c9665b917504e479dcc38ee31b1643a1b7d) with the failing tests. Lets figure out how best to handle them.

### DIFF
--- a/esprima.js
+++ b/esprima.js
@@ -3370,16 +3370,26 @@
     // ECMA-262 12.3.3 The new Operator
 
     function parseNewExpression() {
-        var callee, args, node = new Node();
+        var callee, args, node = new Node(),
+        newNode, newToken, newId,
+        targetNode, targetId;
 
-        expectKeyword('new');
+        newNode = new Node();
+        newToken = lex();
+        if (newToken.type !== Token.Keyword || newToken.value !== 'new') {
+            throwUnexpectedToken(newToken);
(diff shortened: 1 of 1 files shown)


ariya:
There's only a few meta property tests. It's probably quite easy to delete the correspoding baseline fixtures so that they will be generated again.

jasonLaster:
sounds good.

ariya:
@jasonLaster Let's hold this off as this needs to be reimplemented anyway with the new modular code. Please let me know if you need further help, also with the baseline test adjustment.

jasonLaster:
sounds good. sorry for the slow reply

PR #1352, opened 2015-10-29
Use let and const instead of var.

### LINKED ISSUE #1177: Start using the es6
Since esprima soon will have full support for the es6 it makes sense to start using it in the source-code too.

Relates to https://github.com/jquery/esprima/issues/1176

ariya:
Fixes #1177

### DIFF
--- a/src/scanner.ts
+++ b/src/scanner.ts
@@ -87,7 +87,7 @@ export class Scanner {
     };
 
     popState() {
-        var state = this.stateStack.pop();
+        const state = this.stateStack.pop();
         assert(typeof state === 'object', 'Scanner state is not balanced');
 
         this.strict = state.strict;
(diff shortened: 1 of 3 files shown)


src/scanner.ts
```diff
@@ -497,13 +490,11 @@ export class Scanner {
...
-        var start, id, type;
-
-        start = this.index;
+        let type: Token;
+        const start = this.index;
 
         // Backslash (U+005C) starts an escaped character.
-        start = this.index;
-        id = (this.source.charCodeAt(this.index) === 0x5C) ? this.getComplexIdentifier() : this.getIdentifier();
+        const id = (this.source.charCodeAt(start) === 0x5C) ? this.getComplexIdentifier() : this.getIdentifier();
```
mikesherov: Good catch

mikesherov:
LGTM

PR #1354, opened 2015-10-31
Allow CLIs to read from STDIN

gibson042:
Makes the input file optional (defaulting to standard input), and allows `-` to explicitly represent standard input.

Example invocations:
```sh
$ echo '"use\x20strict"' | bin/esparse.js -
{
    "type": "Program",
    "body": [
        {
            "type": "ExpressionStatement",
            "expression": {
                "type": "Literal",
                "value": "use strict",
                "raw": "\"use\\x20strict\""
            }
        }
    ],
    "sourceType": "script"
}
$ bin/esvalidate.js <<EOF
"use strict"
)
EOF
Error: Line 2: Unexpected token )
```

### DIFF
--- a/bin/esparse.js
+++ b/bin/esparse.js
@@ -32,7 +32,7 @@ if (typeof require === 'function') {
     try {
         esprima = require('esprima');
     } catch (e) {
-        esprima = require('../src/esprima');
+        esprima = require('../dist/esprima');
     }
 } else if (typeof load === 'function') {
     try {
--- a/bin/esparse.js
+++ b/bin/esparse.js
@@ -25,7 +25,7 @@
 
 /*jslint sloppy:true node:true rhino:true */
 
-var fs, esprima, fname, content, options, syntax;
+var fs, esprima, fname, forceFile, content, options, syntax;
 
 if (typeof require === 'function') {
     fs = require('fs');
@@ -53,7 +53,7 @@ if (typeof console === 'undefined' && typeof process === 'undefined') {
 
 function showUsage() {
     console.log('Usage:');
-    console.log('   esparse [options] file.js');
+    console.log('   esparse [options] [file.js]');
     console.log();
     console.log('Available options:');
     console.log();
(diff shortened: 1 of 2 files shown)


gibson042:
> I think that just means we have to commit the built file from time to time (at least before a release).

I would not recommend that... Sizzle does so now, and will be moving away from it. The problem is that every build then modifies files under source control. But it turns out that insisting upon a build before local `require` (and therefore CLI use) _is_ common in other tools, and therefore they are implicitly or explicitly using a `dist` analogue. I'll just update this PR as you suggested.

ariya:
Sounds good. While you're at it, I recommend squashing everything into one commit. I doubt that the individual intermediate step is particularly interesting vs the whole change.

gibson042:
I generally like to keep commit history in the PR and leave squashing to maintainers. Would you rather I do it here, though?

bin/esparse.js
```diff
@@ -90,22 +97,14 @@ process.argv.splice(2).forEach(function (entry) {
         options.tokens = true;
     } else if (entry === '--tolerant') {
         options.tolerant = true;
-    } else if (entry.slice(0, 2) === '--') {
+    } else if (entry === '--') {
+        forceFile = true;
```
gibson042: This will cause mishandling of `bin/esparse.js - --`, but that seems rare enough that I didn't consider it worth adding extra code to address the interactions.

bin/esvalidate.js
```diff
@@ -101,17 +112,16 @@ process.argv.splice(2).forEach(function (entry) {
             console.log('Error: unknown report format ' + options.format + '.');
             process.exit(1);
         }
-    } else if (entry.slice(0, 2) === '--') {
+    } else if (entry === '--') {
+        forceFile = true;
```
gibson042: Likewise here, although I suppose this would be little more than `forceFileIndex = fnames.length`. Let me know if you want the robustness.

ariya:
> I generally like to keep commit history in the PR and leave squashing to maintainers. Would you rather I do it here, though?

Whichever easier for you. I can do one or another :-)

gibson042:
Then I leave it to you. Assuming you don't want the more robust `--` handling, this should be good to go.

ariya:
Cool, thanks. And yeah, I agree that robustifying `--` handling is a bit overkill for now.

PR #1361, opened 2015-11-05
Remove the deprecated guardedHandler and handlers of TryStatement.

### LINKED ISSUE #1030: TryStatement should use handler instead of handlers
The `TryStatement` node has changed according to https://github.com/estree/estree/issues/1 from

```js
interface TryStatement <: Statement {
    type: "TryStatement";
    block: BlockStatement;
    handlers: [ CatchClause ];
    guardedHandlers: [];
    finalizer: BlockStatement | null;
}
```

to

```js
interface TryStatement <: Statement {
    type: "TryStatement";
    block: BlockStatement;
    handler: CatchClause | null;
    finalizer: BlockStatement | null;
}
```

The differences are:

 * `handlers` (an array of `CatchClause`) is changed to `handler` (a single `CatchClause`, if any)
 * `guardedHandlers` is gone

See also https://bugzilla.mozilla.org/show_bug.cgi?id=742612.

(Migrated from https://code.google.com/p/esprima/issues/detail?id=433)

ariya:
Fixes #1030


mikesherov:
Does this break webpack? Should I submit a PR to them to fix it?

ariya:
> Does this break webpack? Should I submit a PR to them to fix it?

Yes, but since next major version of webpack won't plan to use Esprima, I therefore pull it out of our downstream test.

mikesherov:
:+1:

PR #1369, opened 2015-11-08
Sauce Labs: Mobile Safari on iPhone emulator of El Capitan.

ariya:


ariya:
This is an attempt to stop the grief on Sauce Labs testing caused by flaky Mobile Safari on iPhone simulator. I don't know the right fix, so let's speculatively update it to running on OS X 10.11 (El Capitan) instead of OS X 10.10 (Yosemite).

ariya:
Success! :sunglasses:

image

PR #1370, opened 2015-11-08
Track strict mode in the parser, vs in the scanner.

### LINKED ISSUE #1368: Track strict mode in the parser
A major part of the most recent refactoring is the unified state tracking in the parser. However, strict mode tracking is still done in the scanner. This is unnecessary since the scanner's concern with strict mode is the so-called "upgrading" of an identifier to a keyword, if it is a strict mode reserved word (e.g. `yield` or `let`).

ariya:
Fixes #1368

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -13,6 +13,21 @@ let PlaceHolders = {
     ArrowParameterPlaceHolder: 'ArrowParameterPlaceHolder'
 };
 
+function upgradeToken(token) {
+    if (token && state.strict && token.type === Token.Identifier) {
+        if (scanner.isStrictModeReservedWord(token.value)) {
+            token.type = Token.Keyword;
+        }
+    }
+    return token;
+}
+
+function nextToken() {
+    const token = scanner.lex();
+    scanner.lookahead = upgradeToken(scanner.lookahead);
+    return token;
+}
+
 function Node(startToken?) {
     if (startToken) {
         if (options.range) {
@@ -770,7 +785,7 @@ function unexpectedTokenError(token?: any, message?: string): Error {
             if (token.type === Token.Keyword) {
                 if (scanner.isFutureReservedWord(token.value)) {
                     msg = Messages.UnexpectedReserved;
-                } else if (scanner.strict && scanner.isStrictModeReservedWord(token.value)) {
+                } else if (state.strict && scanner.isStrictModeReservedWord(token.value)) {
                     msg = Messages.StrictReservedWord;
                 }
             }
@@ -808,7 +823,7 @@ function tolerateUnexpectedToken(token?, message?) {
 // If not, an exception will be thrown.
 
 function expect(value) {
-    const token = scanner.lex();
+    const token = nextToken();
     if (token.type !== Token.Punctuator || token.value !== value) {
         throwUnexpectedToken(token);
     }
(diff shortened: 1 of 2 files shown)


ariya:
I admit that this is just the beginning, as the parser itself can optimize the moment when such a keyword needs to be identifier. In this first step, I just want to remove strict mode tracking from the scanner, i.e. getting it out of the way.

src/parser.ts
```diff
@@ -2412,13 +2427,13 @@ function parseVariableDeclaration(options) {
     const id = parsePattern(params, 'var');
 
     // ECMA-262 12.2.1
-    if (scanner.strict && scanner.isRestrictedWord(id.name)) {
+    if (state.strict && scanner.isRestrictedWord(id.name)) {
```
mikesherov: In a follow up, might be a good idea to turn `state.strict && scanner.isRestrictedWord(token)` into its own function, it's used in enough places.

src/parser.ts
```diff
@@ -13,6 +13,21 @@ let PlaceHolders = {
     ArrowParameterPlaceHolder: 'ArrowParameterPlaceHolder'
 };
 
+function upgradeToken(token) {
+    if (token && state.strict && token.type === Token.Identifier) {
```
mikesherov: Even here, you'd be able to avoid the double nested `if` if we had a function that combined `state.strict && scanner.isStrictModeReservedWord`

src/parser.ts
```diff
@@ -13,6 +13,21 @@ let PlaceHolders = {
     ArrowParameterPlaceHolder: 'ArrowParameterPlaceHolder'
 };
 
+function upgradeToken(token) {
```
mikesherov: An alternative is to make upgradeToken a method of the scanner that receives the strict mode state as an argument. It reads like feature envy in its current form. This function cares about the token, it's type, it's value, its status as a strict mode reserved word, and whether or not the current state is strict. 4 out of 5 of those things are responsibilities of the scanner.

src/parser.ts
```diff
@@ -13,6 +13,21 @@ let PlaceHolders = {
...
+        if (scanner.isStrictModeReservedWord(token.value)) {
+            token.type = Token.Keyword;
+        }
+    }
+    return token;
+}
+
+function nextToken() {
+    const token = scanner.lex();
+    scanner.lookahead = upgradeToken(scanner.lookahead);
```
mikesherov: Even here, the scanner is asking the parser to upgrade the token, which is produced by the scanner, and uses another method of the scanner to make that determination.

src/parser.ts
ariya: In the subsequent improvement, I play with that idea. Remains to be seen whether it is worth it since it won't be called too many times anymore.

src/parser.ts
ariya: That would defeat the point. A scanner should never need to know whether the current parsing happens in the context of strict mode or not. This "upgrade" mechanism is an artifact, a historical baggage that would have not exist if the design was different.

src/parser.ts
ariya: And thus why the goal of the next refactoring is to let the parser works on its own w.r.t strict mode reserved word with the need to taint the scanner at all.

mikesherov: Got it. Then that makes sense.

mikesherov:
:+1:

PR #1371, opened 2015-11-10
Move regular expression literal parsing out of the scanner.

### LINKED ISSUE #1111: Distinguish between scanning regex for parsing vs tokenization

ariya:
Fixes #1111

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -28,6 +28,57 @@ function nextToken() {
     return token;
 }
 
+function nextRegexToken() {
+    let pos, loc;
+
+    if (options.tokens) {
+        scanner.skipComment();
+
+        pos = scanner.index;
+        loc = {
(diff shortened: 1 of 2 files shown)


ariya:
This is similar in spirit as the previous pull request (#1370). The goal is to fix the layer violation, i.e. the scanner should have no business in knowing that the regular expression is being constructed for the parser. More refactoring is still planned to clean-up further mess in the parser itself.

mikesherov:
:+1:

PR #1372, opened 2015-11-10
Fix Program node's sourceType.

ariya:
Somehow this was mistakenly introduced during the TypeScript migration.

Refs #1159


mikesherov:
How was this not caught by a test?

mikesherov:
:+1:

ariya:
> How was this not caught by a test?

Technically speaking, source type never changes. So this is just for strict consistency.

PR #1373, opened 2015-11-11
Use method syntax for function properties

DanielRosenwasser:
While not semantically any different, you can take advantage of the shorter ES6 syntax. I thought it'd be worth a shot to make the change and see if it was desirable.

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -111,7 +111,7 @@ function Node(startToken?) {
 
 Node.prototype = {
 
-    processComment: function() {
+    processComment() {
         let bottomRight = state.bottomRightStack;
         let last = bottomRight[bottomRight.length - 1];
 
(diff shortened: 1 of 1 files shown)


ariya:
Hi @DanielRosenwasser, thanks for the suggestion! Could you please sign the CLA as noted by the above jquery bot?

JFYI, most part of the parser is still under a big refactoring, so usually we try to hold off various syntax tweaks. Also, it's a good idea to file an issue first to potentially discuss your plan ahead of time.

DanielRosenwasser:
Sure; if you're not ready to commit to the change, we can hold off.

I actually used my github noreply email for the change. I then patched up the commit to use my own email address and force-pushed, but it seems like it's still not okay even though I signed the CLA with both emails. Any ideas?

ariya:
Sometimes the CLA checker is flaky.

@DanielRosenwasser In all cases, let's hold this off for now. Once the dust settles a little bit, we will desperately need your deep TypeScript expertise! It would be fantastic if there is an overall review to ensure that we are following idiomatic TypeScript.

DanielRosenwasser:
Sounds great - the team was pretty excited to hear about the recent work on Esprima, so I'll try to help the best I can. :smile:

PR #1375, opened 2015-11-12
Split the state tracking between the scanner and the parser.

### LINKED ISSUE #1374: Revisit state tracking in the parser vs scanner
For a better distribution of responsibilities, the state tracker needs some refactoring.

Token lookahead is necessary to determine the proper grammar production. This needs to exist in the parser as the scanner has no business regarding the production. The same goes for the extraneous location tracking for syntax node's location information.

The implementation of the reader algorithm for the pure tokenizer (i.e. `tokenize` API) shall not be in the parser as well. One day we will separate the tokenizer and find a new home for it, but for now it is sufficient to include it as part of the scanner.

ariya:
Token lookahead, extraneous location tracking for syntax node, etc are
mostly moved to the parser.

Reader algorithm (to determine whether `/` starts a regular expression
or not) for the pure tokenizer now resides completely within the
scanner.

Fixes #1374

### DIFF
--- a/src/scanner.ts
+++ b/src/scanner.ts
@@ -11,6 +11,20 @@ function octalValue(ch: string): number {
     return '01234567'.indexOf(ch);
 }
 
+// A function following one of those tokens is an expression.
+function beforeFunctionExpression(t: string): boolean {
+    return ['(', '{', '[', 'in', 'typeof', 'instanceof', 'new',
+        'return', 'case', 'delete', 'throw', 'void',
+    // assignment operators
+        '=', '+=', '-=', '*=', '/=', '%=', '<<=', '>>=', '>>>=',
+        '&=', '|=', '^=', ',',
+    // binary/unary operators
+        '+', '-', '*', '/', '%', '++', '--', '<<', '>>', '>>>', '&',
+        '|', '^', '!', '~', '&&', '||', '?', ':', '===', '==', '>=',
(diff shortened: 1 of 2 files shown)


src/scanner.ts
```diff
@@ -1258,46 +1202,57 @@ export class Scanner {
...
+                case '}':
+                    // Dividing a function by anything makes little sense,
+                    // but we have to check for that.
+                    regex = false;
+                    if (this.tokenValues[this.openCurlyToken - 3] === 'function') {
+                        // Anonymous function, e.g. function(){} /42
+                        const check = this.tokenValues[this.openCurlyToken - 4];
+                        regex = check ? !beforeFunctionExpression(check) : false;
+                    } else if (this.tokenValues[this.openCurlyToken - 4] === 'function') {
+                        // Named function, e.g. function f(){} /42/
```
mikesherov: Completely unrelated to this PR, but wanted to jot it down before I forgot....

Do we need to add cases for generator functions and arrow functions divided by a number here?

ariya: > Do we need to add cases for generator functions and arrow functions divided by a number here?

Most like the reader algorithm needs an update anyway for ES6. Let's watch sweet.js and when it happens, also adopt the updated algorithm.

ariya:
This is going to land soon so we can continue for further refactoring (yay!).

mikesherov:
:+1:

Issue #1247, opened 2015-07-22
Downstream test: failing JSCS tests


ariya:
This is discovered via #1245 (fails for me with Node.js v0.12.0, npm 2.9.1).

```
git clone https://github.com/jscs-dev/node-jscs.git
cd node-jscs
npm install
npm test
```

ariya:
In the interim, we disable JSCS in the downstream tests.

ariya:
Seems that this is now working again so I'm going to re-enable it.

ariya:
It's failing again, see e.g. https://circleci.com/gh/jquery/esprima/30. Thus I likely will exclude JSCS from the downstream test, at least temporarily.

ariya:
This is becoming a problem again.

ariya:
This happens again, see PR #1448. For the build log: https://circleci.com/gh/jquery/esprima/235. The last few lines are:

```
========== jscs

Cloning https://github.com/jscs-dev/node-jscs.git ...
Cloning into 'jscs'...
remote: Counting objects: 549, done.
remote: Compressing objects: 100% (421/421), done.
remote: Total 549 (delta 189), reused 256 (delta 75), pack-reused 0
Receiving objects: 100% (549/549), 392.86 KiB | 0 bytes/s, done.
Resolving deltas: 100% (189/189), done.
Checking connectivity... done.

HEAD is
commit b19b33e749ee4e715b70f44c5c2ef40569cbaf40
Author: David Harkness <dharkness@gmail.com>
Date:   Wed Jan 6 17:43:02 2016 -0800

    disallowKeywordsOnNewLine: fix documentation format
    
    Two of the examples had an extra space before the comment `*` which messes u
    
    Closes gh-2052

(END)command npm run circleci took more than 1200 seconds since last output
```

@mikesherov Is JCSS test suite known to be reliable when running under Node.js 0.12 (what CircleCI is set to use)?

hzoo:
Yeah haven't seen any issues other than this - cc @markelog?

markelog:
jscs, with all those dependencies is pretty big, like 250M big, 200 of them are dependencies, probably the biggest package of your integration tests and since results are non-deterministic... i think you need to start to use the cache or increase the timeout

ariya:
OMG 250 MB dependencies! I think CircleCI already caches `~/.node_modules`, not sure what's going on there. Perhaps I shall run the `npm install` step in verbose mode to get more details?

ariya:
[...]

PR #1378, opened 2015-11-14
Downstream test: include redeyed.

### LINKED ISSUE #1377: Downstream test: include redeyed
Now that [redeyed](https://github.com/thlorenz/redeyed) is switched to upstream Esprima 2 (https://github.com/thlorenz/redeyed/pull/4), it should be included as part of the downstream project test. Due to its popularity, particularly via [cardinal](https://www.npmjs.com/package/cardinal), it is imperative not to regress redeyed.

ariya:
Closes #1377

### DIFF
--- a/test/downstream.js
+++ b/test/downstream.js
@@ -96,8 +96,9 @@ test_downstream({
     'esmangle': 'https://github.com/estools/esmangle.git',
     'escomplex-js': 'https://github.com/philbooth/escomplex-js.git',
     'js2coffee': 'https://github.com/js2coffee/js2coffee.git',
-    'jscs': 'https://github.com/jscs-dev/node-jscs.git',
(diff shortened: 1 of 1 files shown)


mikesherov:
:+1: congrats on the pickup ;-)

ariya:
Unfortunately we can't verify this yet because CircleCI failed to complete the tests of JSCS. This seems to happen quite often with JSCS. I'm going to experiment with the change of the order so that JSCS becomes the last one.

mikesherov:
I wonder why JSCS fails so often. Ping @hzoo

hzoo:
Doesn't look like you can see any error messages? It just checks out master and then `(END)the build was canceled`? Or maybe I don't understand that output

ariya:
@hzoo Better tracking it at #1247.

PR #1379, opened 2015-11-14
Downstream tests should be executed using the latest Node.js 4.1.

ariya:
This is just a speculative attempt to fix a hit-and-miss issue on CircleCI running various downstream project test suites.

/cc @mikesherov @hzoo

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,7 +1,7 @@
 machine:
   node:
     version:
-      0.12.0
+      4.1.0
   environment:
     NO_WATCH_TESTS: 1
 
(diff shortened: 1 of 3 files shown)


hzoo:
Yay? Not sure what was going on before

ariya:
I'm still struggling to convince CircleCI to ignore the cache (Node.js 0.12 and other npm modules) and thereby to use Node.js 4.1.

ariya:
Too bad Node.js will cause Istanbul tests to fail :-(
Looking at Istanbul's [Travis CI](https://travis-ci.org/gotwarlost/istanbul), seems that it is only tested well with 0.12.

ariya:
@hzoo @mikesherov Same situation, Node.js 4 fails to run JSCS test: https://circleci.com/gh/jquery/esprima/136. You can reproduce this by cloning this repo and run:
```
npm i && npm run downstream
```

I'm closing this since clearly its speculative change doesn't have any meaningful effect.

PR #1380, opened 2015-11-14
Downstream tests should be executed using the io.js.

ariya:
Since Node.js 4 doesn't work (see #1379), let's try something a bit older.


ariya:
Ah, Istanbul tests still won't fly with iojs. I guess I need to try to fix it upstream.

PR #1382, opened 2015-11-15
Track and collect the comments without the callback.

### LINKED ISSUE #1381: Improve comment collection in the scanner
There is no need to transfer the collected comments via the side-channel, i.e. `addComment` callback.

ariya:
Fixes #1381

### DIFF
--- a/test/utils/evaluate-testcase.js
+++ b/test/utils/evaluate-testcase.js
@@ -203,6 +203,15 @@
         expected = JSON.stringify(tokens, null, 4);
 
         try {
+            // Some variations of the options (tolerant mode, to avoid premature error)
+            esprima.tokenize(code, { tolerant: true, comment: false, loc: false, range: false });
+            esprima.tokenize(code, { tolerant: true, comment: false, loc: false, range: true });
+            esprima.tokenize(code, { tolerant: true, comment: false, loc: true,  range: false });
+            esprima.tokenize(code, { tolerant: true, comment: false, loc: true,  range: true });
+            esprima.tokenize(code, { tolerant: true, comment: true,  loc: false, range: false });
+            esprima.tokenize(code, { tolerant: true, comment: true,  loc: false, range: true });
+            esprima.tokenize(code, { tolerant: true, comment: true,  loc: true,  range: false });
+
             list = esprima.tokenize(code, options);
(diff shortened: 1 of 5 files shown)


test/utils/evaluate-testcase.js
```diff
@@ -203,6 +203,16 @@
         expected = JSON.stringify(tokens, null, 4);
 
         try {
+            // Some variations of the options (tolerant mode, to avoid premature error)
+            esprima.tokenize(code, { tolerant: true, comment: false, loc: false, range: false });
+            esprima.tokenize(code, { tolerant: true, comment: false, loc: false, range: true });
+            esprima.tokenize(code, { tolerant: true, comment: false, loc: true,  range: false });
+            esprima.tokenize(code, { tolerant: true, comment: false, loc: false, range: true });
```
mikesherov: `loc: true` here or it's redundant of two lines above.

test/utils/evaluate-testcase.js
```diff
@@ -203,6 +203,16 @@
...
+            esprima.tokenize(code, { tolerant: true, comment: true,  loc: false, range: false });
+            esprima.tokenize(code, { tolerant: true, comment: true,  loc: false, range: true });
+            esprima.tokenize(code, { tolerant: true, comment: true,  loc: true,  range: false });
+            esprima.tokenize(code, { tolerant: true, comment: true,  loc: false, range: true });
```
mikesherov: `loc: true` here or it's redundant of two lines above.

mikesherov:
Should we pull JSCS out until I or @hzoo can figure out what's going on there?

mikesherov:
Aside from the test nits, :+1: really great work here.

ariya:
> Should we pull JSCS out until I or @hzoo can figure out what's going on there?

Good idea, I reopened #1247 to track that.

test/utils/evaluate-testcase.js
ariya: Good catch!

test/utils/evaluate-testcase.js
ariya: Now that I spot it, I realize it's not needed (the real tokenizer test uses that setting).

PR #1384, opened 2015-11-16
update location of esprima.js in package.json "main" field

michaelficarra:


mikesherov:
:+1:

mikesherov:
We should probably release a patch update after this lands.

ariya:
> We should probably release a patch update after this lands.

Why is that? There is no release yet based on master.

mikesherov:
Ah, I forgot.

PR #1385, opened 2015-11-17
Fix assignment to `let` in non-strict mode.

### LINKED ISSUE #1376: Bug about use future reserved word as identifier
loose mode
```javascript
> require('esprima').parse('let = 1')
Error: Line 1: Invalid left-hand side in assignment
```
Expected: a syntax tree
Actual: an exception

ariya:
Fix #1376


ariya:
Obviously, I will also forward-port this to the master branch.

mikesherov:
:+1:

PR #1386, opened 2015-11-17
Implement MetaProperty identifiers.

### LINKED ISSUE #1338: MetaProperty implemented incorrectly

ariya:
Based on the previous work of @jasonLaster. 

Fixes #1338

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -1807,13 +1807,18 @@ function parseComputedMember() {
 function parseNewExpression() {
     const node = new Node();
 
-    expectKeyword('new');
+    const newNode = new Node();
+    const newToken = nextToken();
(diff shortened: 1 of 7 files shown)


mikesherov:
:+1:

jasonLaster:
:+1: Thanks

Issue #1392, opened 2015-11-21
Downstream test: failing Istanbul tests


ariya:
From pull request #1391:

```
test-browser-instrumentation.js
Start server
ENOENT, no such file or directory '/tmp/downstream1151021-12524-1e7mits/istanbul/node_modules/escodegen/escodegen.browser.min.js'
Error: ENOENT, no such file or directory '/tmp/downstream1151021-12524-1e7mits/istanbul/node_modules/escodegen/escodegen.browser.min.js'
    at Error (native)
    at Object.fs.openSync (fs.js:502:18)
    at Object.fs.readFileSync (fs.js:354:15)
    at reader (/tmp/downstream1151021-12524-1e7mits/istanbul/test/browser/support/server.js:11:19)
    at Object.<anonymous> (/tmp/downstream1151021-12524-1e7mits/istanbul/test/browser/support/server.js:14:23)
    at Module._compile (module.js:460:26)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Module.require (module.js:365:17)
✖ should produce identical instrumentation results when instrumentation run on browser v/s nodejs

Assertion Message: ENOENT, no such file or directory '/tmp/downstream1151021-12524-1e7mits/istanbul/node_modules/escodegen/escodegen.browser.min.js'
AssertionError: ENOENT, no such file or directory '/tmp/downstream1151021-12524-1e7mits/istanbul/node_modules/escodegen/escodegen.browser.min.js'
    at Object.ok (/tmp/downstream1151021-12524-1e7mits/istanbul/node_modules/nodeunit/lib/types.js:83:39)
    at Object.module.exports.should produce identical instrumentation results when instrumentation run on browser v/s nodejs (/tmp/downstream1151021-12524-1e7mits/istanbul/test/browser/test-browser-instrumentation.js:88:18)
    at /tmp/downstream1151021-12524-1e7mits/istanbul/node_modules/nodeunit/lib/core.js:232:20
    at /tmp/downstream1151021-12524-1e7mits/istanbul/node_modules/nodeunit/deps/async.js:168:13
    at /tmp/downstream1151021-12524-1e7mits/istanbul/node_modules/nodeunit/deps/async.js:131:25
    at /tmp/downstream1151021-12524-1e7mits/istanbul/node_modules/nodeunit/deps/async.js:165:17
[...]

PR #1393, opened 2015-11-22
Downstream test: temporarily disable troubling Istanbul test.

ariya:
Refs #1392


ariya:
That seems to do the trick! I'm going to land this workaround till we get a real fix for #1392.

PR #1394, opened 2015-11-22
Extract the tokenizer from parser.ts and put in the new tokenizer.ts.

### LINKED ISSUE #1389: Refactor the pure tokenizer
Rather than being interleaved with other parsing code inside `parser.ts`, the pure tokenizer should reside in its own dedicated source file, `tokenizer.ts`.

ariya:
Fixes #1389

### DIFF
--- a/src/scanner.ts
+++ b/src/scanner.ts
@@ -12,19 +12,6 @@ function octalValue(ch: string): number {
     return '01234567'.indexOf(ch);
 }
 
-// A function following one of those tokens is an expression.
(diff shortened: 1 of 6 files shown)


src/scanner.ts
```diff
@@ -1266,57 +1242,4 @@ export class Scanner {
         return this.scanPunctuator();
     };
 
-    advance() {
```
ariya: This is sweet.js's reader algorithm implementation and it's entirely moved to the Tokenizer class.

mikesherov:
:+1:

PR #1398, opened 2015-11-24
Include the tokenization in the benchmarks.

### LINKED ISSUE #1396: Tokenizer benchmark
Currently `test/benchmarks.js` only exercises the parser. We need to include the tokenizer so that we can be aware of any possible speed regression.

ariya:
Fixes #1396

### DIFF
--- a/test/benchmarks.js
+++ b/test/benchmarks.js
@@ -50,7 +50,7 @@ function kb(bytes) {
     return (bytes / 1024).toFixed(1);
 }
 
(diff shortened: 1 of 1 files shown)


ariya:
Since it's so useful, I also plan to backport this into 2.x. It should be pretty harmless since it matters only when the benchmark is invoked explicitly during the development.

mikesherov:
:+1:

PR #1405, opened 2015-12-01
Reduce duplicated code to handle directives

### LINKED ISSUE #1404: Reduce duplicated code to handle directives
The code to check for directive prologues exists both in `parseFunctionSourceElements()` and `parseScriptBody()`. This is essentially the same and it is better to refactor it into a function that is called from both places.

ariya:
In addition, parse only expression statements as each directive consists
of a string literal and thereby guarantees that it can't be other types
of statements.

Fixes #1404


mikesherov:
:+1:

PR #1408, opened 2015-12-03
Fix missing delegate property in YieldExpression

### LINKED ISSUE #1407: Missing delegate property in a yield expression
When `yield` is followed by a line terminator, somehow the `delegate` property in the syntax node is missing.

Example code to be parsed:
```js
function *g() { yield
}
```

Actual:
```js
> require('esprima').parse('function *g() { yield\n }').body[0].body.body[0]
{ type: 'ExpressionStatement',
  expression: { type: 'YieldExpression', argument: null, delegate: undefined } }
```  

Expected:
```js
> require('esprima').parse('function *g() { yield\n }').body[0].body.body[0]
{ type: 'ExpressionStatement',
  expression: { type: 'YieldExpression', argument: null, delegate: false } }
```

ariya:
Fixes #1407

### DIFF
--- a/esprima.js
+++ b/esprima.js
@@ -3854,6 +3854,7 @@
 
         argument = null;
         expr = new Node();
+        delegate = false;
 
         expectKeyword('yield');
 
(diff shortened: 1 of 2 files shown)


mikesherov:
:+1:

PR #1416, opened 2015-12-08
Construct a MethodDefinition node without mutating another node.

### LINKED ISSUE #1414: Consistent construction of a method definition node
Currently, a `MethodDefinition` node (see `parseClassElement`) is constructed by mutating an existing `Property` node. It is better to have a special constructor for this node (just like others in `nodes.ts`) and have the node created the same consistent way.

ariya:
Fixes #1414

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -913,11 +913,13 @@ class Parser {
         return key;
     }
 
-    checkDuplicatedProto(key, hasProto) {
-        const proto = (key.type === Syntax.Identifier && key.name === '__proto__') ||
-            (key.type === Syntax.Literal && key.value === '__proto__');
+    isPropertyKey(key, value) {
+        return (key.type === Syntax.Identifier && key.name === value) ||
+            (key.type === Syntax.Literal && key.value === value);
+    }
 
-        if (proto) {
+    checkDuplicatedProto(key, hasProto) {
+        if (this.isPropertyKey(key, '__proto__')) {
             if (hasProto.value) {
                 this.tolerateError(Messages.DuplicateProtoProperty);
             } else {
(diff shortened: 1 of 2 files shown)


src/parser.ts
```diff
@@ -3115,14 +3119,12 @@ class Parser {
...
-        property.static = isStatic;
-        if (property.kind === 'init') {
-            property.kind = 'method';
+        if (kind === 'init') {
+            kind = 'method';
         }
         if (!isStatic) {
-            if (!property.computed && (property.key.name || property.key.value.toString()) === 'constructor') {
-                if (property.kind !== 'method' || !property.method || property.value.generator) {
+            if (!computed && this.isPropertyKey(key, 'constructor')) {
```
mikesherov: is there any test you can add that checks the `(key.type === Syntax.Literal && key.value === value)` part of `this.isProperty` for this condition?

src/parser.ts
```diff
@@ -3130,18 +3132,15 @@ class Parser {
                 } else {
                     hasConstructor.value = true;
                 }
-                property.kind = 'constructor';
+                kind = 'constructor';
             }
         } else {
-            if (!property.computed && (property.key.name || property.key.value.toString()) === 'prototype') {
+            if (!computed && this.isPropertyKey(key, 'prototype')) {
```
mikesherov: is there any test you can add that checks the `(key.type === Syntax.Literal && key.value === value)` part of `this.isProperty` for this condition?

mikesherov: can you collapse the `else { if {` into `else if {` please?

mikesherov:
Besides the nits... :+1:

src/parser.ts
ariya: Good point. And now that I've taken a second look, I think the conditional block needs be based on `!computed` anyway for improved readability.

src/parser.ts
ariya: Actually it doesn't count as a constructor and thus this case should be excluded.

PR #1419, opened 2015-12-09
Use UpdateExpression node for postfix and increment/decrement.

### LINKED ISSUE #1417: Use UpdateExpression for a postfix expression and increment/decrement unary
The `UpdateExpression` node needs to be used for two cases:

* In a postfix expression (because ESTree doesn't have the notion of `PostfixExpression`)
* A unary expression with `++` or `--` operator

This is related to the constructors of the AST and not with the actual syntax nodes produced

ariya:
Fixes #1417


mikesherov:
:+1:

PR #1421, opened 2015-12-09
Expression type should include Identifier.

### LINKED ISSUE #1420: type Expression should include Identifier
The `Expression` type in nodes.ts does not include `Identifier`, but in `parsePrimaryExpression` in parse.ts the `expr: Expression` variable is set to an `Identifier` in many code paths. I think it makes sense to classify `Identifier` as an `Expression` because an `ExpressionStatement`'s `expression` property can just be set to an `Identifier`.

The reason TypeScript doesn't complain is because `Expression` includes `ThisExpression` which only has a `type` property defined on it, making it a super-type of `Identifier` (as least as far as TypeScript's duck-typing is concerned).

Incidentally, this means TypeScript will consider *any* node a valid sub-type of `Expression`. One possible workaround (though ugly) could be to add some breaking property to `ThisExpression` like `__isThisExpression__ = true`. I'm not sure what trade-offs are right for Esprima... just pointing it out.

ariya:
Fixes #1420


mikesherov:
:+1:

Issue #1410, opened 2015-12-04
Overhaul the parser


ariya:
Now that the pure tokenizer (#1389), common error handler (#1390), and AST constructors (#1410) are out of `parser.ts`, it is time to clean up the parser code. The first step is to turn the bulk of it into its own class, `Parser`. For each parsing function, the following steps need to be carried out:

  * Use `const` as much as possible inside the function body
  * Have only one return statement at the end (to reduce complexity)
  * Provide the type information for all function paramaters
  * Provide the type information for the return value
  * Eliminate outdated micro-optimizations (watch for performance regressions!)
  * Remove the need to pass around constructed node

PR #1423, opened 2015-12-10
Missing type information for the return value

ariya:
Refs #1410


ariya:
This is almost trivial so it will be going in soon!

PR #1425, opened 2015-12-11
Eliminate the post-processing step of token location filtering

### LINKED ISSUE #1418: Eliminate the post-processing step of token location filtering
In case the list of all tokens needs to be included in the parsing result, the extra step of filtering out the token location information is not needed anymore. With most of the state tracking between the parser vs the scanner is well isolated (see #1374), the token location information can be added on the fly, as needed, rather than as a post-processing step.

ariya:
Fixes #1418

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -274,33 +274,6 @@ class Parser {
         }
     }
 
-    filterTokenLocation(tokens) {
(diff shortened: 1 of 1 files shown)


mikesherov:
:+1:

PR #1426, opened 2015-12-11
Dependencies: update JSCS and TSLint

ariya:


ariya:
This isn't pretty non controversial so I'll merge it.

PR #1428, opened 2015-12-12
Move the exported entry points to the main source file

### LINKED ISSUE #1427: Move the exported entry points to the main source file
With both the `Parser` and `Tokenizer` classes becoming self-contained, now it is time to move the exported functions, `parse` and `tokenize`, into `esprima.ts`.

ariya:
This keeps `parser.ts` and `tokenizer.ts` only contain the respective
class and nothing else.

Fixes #1427


ariya:
Yet another non-controversial-moving-around so it'll likely go in pretty soon.

Issue #1045, opened 2015-02-16
ES6 feature: Destructuring: Tracking Issue


ariya:
# [12.14.5](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-destructuring-assignment) Destructuring Assignment 

They are used in the left hand side of an assignment expression when `=` is the assignment operator, or at the `LeftHandSideExpression` position of a for-in or a for-of loop. 

Referred to by:
* [12.14.1](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-assignment-operators-static-semantics-early-errors) Static Semantics: Early Errors
* [13.6.4.1](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-for-in-and-for-of-statements-static-semantics-early-errors) Static Semantics: Early Errors

## Syntax:

<pre>
<i>AssignmentPattern</i>[Yield] :
  <i>ObjectAssignmentPattern</i><sub>[?Yield]</sub>
  <i>ArrayAssignmentPattern</i><sub>[?Yield]</sub>

<i>ObjectAssignmentPattern</i>[Yield] :
  <b>{</b> <b>}</b>
  <b>{</b> <i>AssignmentPropertyList</i><sub>[?Yield]</sub> <b>}</b>
  <b>{</b> <i>AssignmentPropertyList</i><sub>[?Yield]</sub> <b>,</b> <b>}</b>

<i>ArrayAssignmentPattern</i>[Yield] :
  <b>[</b> <i>Elision</i><sub>opt</sub> <i>AssignmentRestElement</i><sub>[?Yield]</sub><sub>opt</sub> <b>]</b>
  <b>[</b> <i>AssignmentElementList</i><sub>[?Yield]</sub> <b>]</b>
  <b>[</b> <i>AssignmentElementList</i><sub>[?Yield]</sub> , <i>Elision</i><sub>opt</sub> AssignmentRestElement</i><sub>[?Yield]</sub><sub>opt</sub> <b>]</b>

<i>AssignmentPropertyList</i>[Yield] :
  <i>AssignmentProperty</i><sub>[?Yield]</sub>
  <i>AssignmentPropertyList</i><sub>[?Yield]</sub> <b>,</b> <i>AssignmentProperty</i><sub>[?Yield]</sub>

<i>AssignmentElementList</i>[Yield] :
  <i>AssignmentElisionElement</i><sub>[?Yield]</sub>
[...]

PR #1432, opened 2015-12-15
Remove unused counter to track parenthesis

ariya:
This is not necessary anymore once the cover grammar was properly
supported (part of #1045).


ariya:
This is rather harmless, it won't cause regression; thus I like to land it at some point :-)

mikesherov:
:+1:

ariya:
Merged into 2.x branch!

PR #1433, opened 2015-12-17
Downstream test: include JSCS.

ariya:
Speculative attempt to see if #1247 is resolved.


ariya:
Looks like JSCS stops misbehaving @mikesherov @hzoo

hzoo:
Hmm still not sure what changed or what is flaky

PR #1435, opened 2015-12-19
Add the support for syntax node delegator

### LINKED ISSUE #1113: API: Syntax delegator
A sophisticated consumer ("power user") of Esprima might want to know and carry out a certain task when a particular event occurs. Among others, this could be when:

  * a syntax node is completed
  * a comment is encountered

This can be achieved if such a task is delegated to a specified delegator.

For a reference, see also [Delegation Pattern](http://best-practice-software-engineering.ifs.tuwien.ac.at/patterns/delegation.html).

ariya:
The third parameter to `parse()` is a callback function that is invoked
every time the parser encounters and completes a new syntax node.

An example that prints "Literal ExpressionStatement Program":

```js
  var list = [];
  esprima.parse('42', {}, function (node) {
    list.push(node.type);
  });
  console.log(list.join(' '));
```

Fixes #1113

### DIFF
--- a/test/api-tests.js
+++ b/test/api-tests.js
@@ -126,10 +126,12 @@ describe('esprima.parse', function () {
         });
     });
 
-    it('should ignore a third parameter', function () {
-        var random = { p: 1, q: 2 };
+    it('should accept three parameters (source, options, and delegate)', function () {
+        var options = { range: false };
         assert.doesNotThrow(function () {
-            esprima.parse('var x', {}, random);
+            esprima.parse('var x', options, function f(entry) {
+                return entry;
+            });
         });
     });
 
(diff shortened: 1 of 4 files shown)


test/api-tests.js
```diff
@@ -218,6 +220,122 @@ describe('esprima.parse', function () {
...
+    });
+
+    it('should be affected by cover grammars', function () {
+        var list = [];
+        esprima.parse('(x, y) => z', {}, function (n) {
+            list.push(n.type)
+        });
+
+        // (x, y) will be a SequenceExpression first before being reinterpreted as
+        // the formal parameter list for an ArrowFunctionExpression
```
gibson042: Isn't this a _bad_ thing for practical use?

ariya: Depending on the use case. If I use the delegate to perform syntax highlighting, it should be pretty harmless since the later arrow function expression overrides the "covered" sequence expression. If however I use the delegate for a static analyzer that prohibits any use of sequence expressions, then in some cases I need to delay the decision accordingly.

In all cases, if the callback logic can not handle with this sequence of events (pun intended!), then it is not suitable for an in-flight processing and there is no choice but to reimplement it as a post-parsing tree traversal.

gibson042: Understood, but more specifically this feels to me like a leaky abstraction of the underlying implementation. And if so, then what is the value of asserting it with unit tests?

ariya: Good point on the leaky abstraction.
It is unlikely however that the implementation thereof will change,
particularly since cover grammars are part of the specification.
I think there is a benefit of keeping it in the tests as a form of
quasi-documentation.

On Sat, Dec 19, 2015, 13:03 Richard Gibson <notifications@github.com> wrote:

> In test/api-tests.js
> <https://github.com/jquery/esprima/pull/1435#discussion_r48096657>:
>
> > +        var ast = esprima.parse('42', { range: true }, function (node) {
> > +            node.start = node.range[0];
> > +            node.end = node.range[1];
> > +        });
> > +        assert.deepEqual(ast.start, 0);
> > +        assert.deepEqual(ast.end, 2);
> > +    });
> > +
> > +    it('should be affected by cover grammars', function () {
> > +        var list = [];
> > +        esprima.parse('(x, y) => z', {}, function (n) {
> > +            list.push(n.type)
> > +        });
> > +
> > +        // (x, y) will be a SequenceExpression first before being reinterpreted as
> > +        // the formal parameter list for an ArrowFunctionExpression
>
> Understood, but more specifically this feels to me like a leaky
> abstraction of the underlying implementation. And if so, then what is the
> value of asserting it with unit tests?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/jquery/esprima/pull/1435/files#r48096657>.
>
-- 
Ariya Hidayat, twitter.com/AriyaHidayat

gibson042: > I think there is a benefit of keeping it in the tests as a form of quasi-documentation.

Yeah, I can agree to that too.

ariya:
I'm going to merge this soon unless there are other polishes needed.

mikesherov:
:+1:

PR #1439, opened 2015-12-24
Reimplement comment handling using the parser delegate

### LINKED ISSUE #1434: Reimplement comment handling using the delegate
Once the parser delegate is in place (#1113), the code to collect the comments and also to attach them the appropriate node shall be moved out of the Parser class.

ariya:
This moves the bulk of comment handling out of the Parser class.

Fixes #1434

### DIFF
--- a/src/comment-handler.ts
+++ b/src/comment-handler.ts
@@ -1,0 +1,180 @@
+import { Syntax } from './syntax';
+
+interface Comment {
+    type: string;
+    value: string;
+    range?: any;
+    loc?: any;
+}
+
+interface Entry {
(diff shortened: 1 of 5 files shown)


ariya:
This is the logical next step after #1435. With this, the parser gets smaller (again) and becomes less cluttered. A rather minor drawback is a performance regression, although this gets only ~11% in a pathological situation (very large input sprinkled with comments all over the place). I believe this is a small price to pay for a cleaner and more modular code base.

mikesherov:
:+1: Makes sense.

PR #1446, opened 2016-01-08
Travis CI: No need to update npm.

ariya:
Speculative fix to see if this solves the false alarm when running the tests with Node.js 5 on Travis CI.


ariya:
This seems to do the trick!

mikesherov:
:+1:

PR #1448, opened 2016-01-10
Dependencies: update JSCS

ariya:


ariya:
Ironically, the downstream tests failed because JSCS test did not finish...

ariya:
JSCS is failing again, thus I reopen #1247.

PR #1449, opened 2016-01-11
Tolerate unterminated if/while/for

### LINKED ISSUE #1042: Tolerate missing ) for if/for/while
Consider the following test code (parsing in tolerant mode):

```js
require('./esprima').parse('if(foo.bar === 1', { tolerant: true })
```

The current master branch will thrown an exception due to the error `Line 1: Unexpected end of input`.

If this feature is implemented, it will return an AST because the missing `)` is tolerated. The error will be still listed in the `errors` array.

(Migrated from https://code.google.com/p/esprima/issues/detail?id=553, as reported by @mrennie)

ariya:
Fixes #1042


michaelficarra:
This should use `EmptyStatement` nodes instead of `null` so valid estree ASTs are returned.

ariya:
> This should use EmptyStatement nodes...

Good point, I'll change it.

PR #1455, opened 2016-01-15
Use assignment expression for default parameter value

### LINKED ISSUE #1081: Use AssignmentExpression for default parameter value

ariya:
Fixes #1081


mikesherov:
Woah!

mikesherov:
:+1:

ariya:
This weekend is probably going to be a merge-a-thon for me!

PR #1458, opened 2016-01-29
Fix shorthand object destructuring defaults in variable declarations

zjmiller:
In a variable declaration, in an object destructuring pattern, a shorthand property with a default value was being incorrectly parsed as having a false shorthand field. For example, the shorthand property in `var {a = b} = c;` was being incorrectly parsed as having a false shorthand field.

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -1804,6 +1804,7 @@ export class Parser {
             const id = this.parseVariableIdentifier();
             if (this.match('=')) {
                 params.push(keyToken);
+                shorthand = true;
                 this.nextToken();
                 key = id;
                 const expr = this.parseAssignmentExpression();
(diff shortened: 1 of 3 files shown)


ariya:
@zjmiller Was there a bug filed regarding this?

zjmiller:
I don't think so.

zjmiller:
I just noticed Acorn and Esprima were giving different ASTs for these cases and looked into it.

zjmiller:
Also this isn't an issue in assignment expressions. Parsing an assignment expression involves `parseObjectProperty` (which does set the shorthand field to true). Parsing a variable declaration, on the other hand, involves `parsePropertyPattern` (which was failing to set the shorthand field to true).

ariya:
Can you file a bug? It's just easier for future reference when there is a summary and an example case. Follow the format as given in past bugs, e.g. #1223.

zjmiller:
Sure.

ariya:
The failing CircleCI tests is because one of our downstream projects, JSCS, is not using Esprima anymore /cc @mikesherov. I'll remove it from test soon.

ariya:
I'll merge this once the updated downstream tests (#1460) doesn't show any problem.

PR #1460, opened 2016-01-30
Downstream update: remove JSCS, add Rocambole and Recast

ariya:


ariya:
The run time for the complete downstream project test on CircleCI does not change (~16 minutes).

PR #1462, opened 2016-02-01
Fix out-of-bound error location in an invalid string literal

### LINKED ISSUE #1457: Invalid position reported after scanStringLiteral hits bad code
It appears that in the `scanStringLiteral()` function [here](https://github.com/jquery/esprima/blob/2.7.1/esprima.js#L1056) when scanning an invalid "string" that doesn't finish before the end of the line, the position reported as the invalid character is incorrect.

What it looks like is going on is when it hits the line ending and `break;`s [here](https://github.com/jquery/esprima/blob/2.7.1/esprima.js#L1132) the current index will have been incremented past the end of the line [here](https://github.com/jquery/esprima/blob/2.7.1/esprima.js#L1067) and is never corrected since it has broken out of the loop and is in the process of calling `throwUnexpectedToken()`

When linting this code:

```JavaScript
var testError = image/png'
```

This results in incorrect results when a linter is using `esprima` for parsing:
image

Originally filed over [here](https://github.com/jscs-dev/node-jscs/issues/2083), with reports that seem to be related to this coming in from users of `jscs` and `jshint` from the Atom linter plugins.

Version: `esprima` v2.7.1
Operating System: Windows 10 x64

If you need any further information I'd be happy to help.

ariya:
When a string literal is not valid, mark the initial quote as the error
location since it is considered the offending (illegal) token.

Fixes #1457

### DIFF
--- a/src/scanner.ts
+++ b/src/scanner.ts
@@ -882,6 +882,7 @@ export class Scanner {
         }
 
         if (quote !== '') {
+            this.index = start;
             this.throwUnexpectedToken();
         }
 
--- a/test/fixtures/invalid-syntax/migrated_0031.failure.json
+++ b/test/fixtures/invalid-syntax/migrated_0031.failure.json
@@ -1,6 +1,6 @@
 {
-    "index": 7,
+    "index": 0,
     "lineNumber": 1,
-    "column": 8,
+    "column": 1,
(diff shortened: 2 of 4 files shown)


mikesherov:
:+1:

PR #1473, opened 2016-02-18
Add boolean and number to Literal's value property

### LINKED ISSUE #1471: Class Literal#value should allow boolean and number types?
The `value` property on class `Literal` is currently [typed as `string` only](https://github.com/jquery/esprima/blob/master/src/nodes.ts#L447). Shouldn't it be `boolean|number|string`?

ariya:
Fixes #1471


mikesherov:
:+1:

PR #1489, opened 2016-03-24
fixes #1487: defend against getters starting with $ on Object.prototype

### LINKED ISSUE #1487: Can't use 'type' as the name of any function variable. (Under special circumstances)
Check it out trying to parse

    (function(a, b, c) { console.log('WOO'); })() // WORKS

but

    (function(a, b, type) { console.log('SUCKS'); })() // DOESN'T

Using esprima on node, fresh install, `npm install esprima` and didn't mess with anything.

By the way, strict mode is broken but I guess that's 'expected behavior'.

michaelficarra:
Fixes #1487.


michaelficarra:
CircleCI fails because of redeyed...

ariya:
:+1: Thanks!

PR #1491, opened 2016-04-17
Dependencies: update jscs and tslint

ariya:


test/check-version.js
```diff
@@ -26,13 +26,16 @@
 
 'use strict';
 
-var fs = require('fs');
+function readFile(filename) {
```
ariya: A minor harmless refactoring, also to satisfy JSCS requireVarDeclFirst rule.

ariya:
False alarm on the failing CircleCI test: I thought I fixed the troublesome redeyed downstream test (https://github.com/thlorenz/redeyed/pull/6), but apparently it is still out of control 😭

PR #1495, opened 2016-06-06
deep clone node when node.shorthand = true

fishbar:
case 1:
```
let name = 'test';
let email = ''
let user = {name, email}
```

case 2:

```
let user = {name: 'test', email: 'abc@xxx.com'}
let {name, email} = user;
```

now esprima parse js code into ast as following:
```
{
    "type": "ObjectExpression",
    "properties": [
        {
            "type": "Property",
            "key": {
                "type": "Identifier",
                "name": "name"
            },
            "computed": false,
            "value": {
                "type": "Identifier",
                "name": "name"
            },
            "kind": "init",
            "method": false,
            "shorthand": true
        },
        {
            "type": "Property",
            "key": {
                "type": "Identifier",
                "name": "email"
            },
            "computed": false,
            "value": {
                "type": "Identifier",
                "name": "email"
            },
            "kind": "init",
            "method": false,
            "shorthand": true
        }
    ]
}
```

each property, the  `key` ===  the `value` , so when eshortten change the Identifier name into shortten chars,  bug cames:
```
let {a, b} = user;
```
=====================

this pr deep clone the value object, so the `key` !== `value`, and bug resolved.

ps: the deepClone method should overwrite by typescript, XD

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -42,6 +42,26 @@ interface MetaNode {
     column: number;
 }
 
+function deepClone(obj) {
+    let res = {}, prop, value;
+    if (!obj) {
+        return obj;
+    }
+
+    for (prop in obj) {
+        if (!obj.hasOwnProperty(prop)) {
+            continue;
+        }
+        value = obj[prop];
+        if (typeof value === 'object') {
+            res[prop] = deepClone(value);
+        } else {
+            res[prop] = value;
+        }
+    }
+    return res;
+}
+
 const ArrowParameterPlaceHolder = 'ArrowParameterPlaceHolder';
 
 interface ArrowParameterPlaceHolderNode {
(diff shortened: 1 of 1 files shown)


ariya:
Thanks @fishbar! Could you also report an issue as described in our [Filing a Defect](https://github.com/jquery/esprima/wiki/Filing-a-Defect) guide?

michaelficarra:
We don't need (and shouldn't have) a full deep clone implementation. We know ahead of time what the structure of the object is.

fishbar:
@ariya  ok, later i post one.

@michaelficarra  so did you means i should clone it by hardcode, not implement an clone function ?

michaelficarra:
Yes, just construct a new node of the expected type.

ariya:
Superceded by PR #1520.

PR #1513, opened 2016-08-16
Ignore the catch clause when testing regular expressions

### LINKED ISSUE #1512: Coverage regression on Node.js v6
1. Use Node.js >= 6.1
2. Clone this repository
3. Run `npm i && npm t`

**Actual**:
```
ERROR: Coverage for statements (99.97%) does not meet global threshold (100%)
```

**Expected**: No error.

ariya:
In a JavaScript environment that supports all regular expression flags,
the catch clause has no effect and thus shall be ignored.

Fixes #1512

### DIFF
--- a/src/scanner.ts
+++ b/src/scanner.ts
@@ -1072,6 +1072,7 @@ export class Scanner {
         try {
             return new RegExp(pattern, flags);
         } catch (exception) {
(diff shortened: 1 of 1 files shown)


michaelficarra:
LGTM.

PR #1518, opened 2016-08-22
Release 2.7.3

ariya:
Let's do another patch update on the 2.7 series to include one tokenizer fix and to ensure that everything runs on Node.js v6!


ariya:
Tagged, published to npm, and used in the web site!

PR #1520, opened 2016-08-24
Ensure that value object in literal property shorthand is deep copied

### LINKED ISSUE #1519: Value object in literal property shorthand should not be a reference
Test case:

```javascript
var program = esprima.parse('x = {y, z}');
var object = program.body[0].expression.right;
object.properties[0].key.name = 'a';
console.log(object.properties[0].value);
```

Actual output:
```
Identifier { type: 'Identifier', name: 'a' }
```

Expected:
```
Identifier { type: 'Identifier', name: 'y' }
```

ariya:
Fixes #1519

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -864,8 +864,8 @@ export class Parser {
                 method = true;
 
             } else if (token.type === Token.Identifier) {
+                const id = this.finalize(node, new Node.Identifier(token.value));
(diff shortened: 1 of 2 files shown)


mikesherov:
:+1:

ariya:
For completeness, I'll backport this to the 2.x branch as well.

PR #1529, opened 2016-08-26
Fix the issue with primary expression in JSX attribute

### LINKED ISSUE #1526: Lex/parse failure when checking for empty {} in JSX attributes
Example:
```
  ReactDOM.render(
    <CommentBox url="/api/comments" pollInterval={2000} />,
    document.getElementById('content')
  );
```
The problem occurs in jsx-parser.ts around line 316:
```
        this.expectJSX('{');
        var expression = null;
        if (this.matchJSX('}')) {
            this.tolerateError('JSX attributes must only be assigned a non-empty expression');
        }
        else {
            this.finishJSX();
            expression = this.parseAssignmentExpression();
            this.startJSX();
        }
        this.expectJSX('}');
```
The problem is that the matchJSX('}') call is executed with the scanner still in JSX mode, so the expression between {} will fail if it starts with any token that is not also a JSX token. I don't know the code well enough yet to suggest a proper fix, but it's definitely not safe to call matchJSX() at that point.

ariya:
Fixes #1526


mikesherov:
:+1:

ariya:
Thanks @mikesherov! We are soon back to zero JSX problem, yay 😸

PR #1531, opened 2016-08-26
Cross-browser tests: temporarily exclude Internet Explorer 9

ariya:
From the most recent Sauce Labs runs, IE 9 failed to launch.
It is either the problem with IE 9 or Windows 7 or both.


mikesherov:
:+1:

ariya:
Merged!

PR #1532, opened 2016-08-27
Fix the regression with the default parameter value

### LINKED ISSUE #1508: Where is a default value?
```js
esprima.parse('let fn = (p1=1) => p1 * p1')
```
returns:
```json
{
   "type":"Program",
   "body":[
      {
         "type":"VariableDeclaration",
         "declarations":[
            {
               "type":"VariableDeclarator",
               "id":{
                  "type":"Identifier",
                  "name":"fn"
               },
               "init":{
                  "type":"ArrowFunctionExpression",
                  "id":null,
                  "params":[
                     {
                        "type":"Identifier",
                        "name":"p1"
                     }
                  ],
                  "body":{
                     "type":"BinaryExpression",
                     "operator":"*",
                     "left":{
                        "type":"Identifier",
                        "name":"p1"
                     },
                     "right":{
                        "type":"Identifier",
                        "name":"p1"
                     }
                  },
                  "generator":false,
                  "expression":true
               }
            }
         ],
         "kind":"let"
      }
   ],
   "sourceType":"script"
}
```
Why syntax tree have not a default value for `p1`?

`esprima v3`

But it exists if i put same code to http://esprima.org/demo/parse.html

ariya:
This was introduced in commit f780eef307 when trying to adhere to the latest
ESTree specification on default parameter value.

Fixes #1508

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -1483,25 +1483,19 @@ export class Parser {
 
         for (let i = 0; i < params.length; ++i) {
             const param = params[i];
-            switch (param.type) {
-                case Syntax.AssignmentPattern:
-                    params[i] = param.left;
-                    if (param.right.type === Syntax.YieldExpression) {
-                        if (param.right.argument) {
-                            this.throwUnexpectedToken(this.lookahead);
-                        }
-                        param.right.type = Syntax.Identifier;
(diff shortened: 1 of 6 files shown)


mikesherov:
:+1:

PR #1540, opened 2016-09-03
Simplfy unary expression parsing

### LINKED ISSUE #1539: Simplify unary expression parsing logic
A side effect of exponentiation expression support (#1490) is that the prefix increment (with `--` and `++`) is not part of _UnaryExpression_ anymore. This also means that `parseUnaryExpression()` can be simplified as well.

ariya:
Fixes #1539

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -1304,15 +1304,8 @@ export class Parser {
     parseUnaryExpression(): Node.Expression {
         let expr;
 
-        if (this.match('+') || this.match('-') || this.match('~') || this.match('!')) {
-            const node = this.startNode(this.lookahead);
-            const token = this.nextToken();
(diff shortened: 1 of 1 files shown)


mikesherov:
Super tiny micro-optimization might be switching `this.context.strict && expr.operator === 'delete'` to `expr.operator === 'delete' && this.context.strict` considering more tokens will lead to this check now and the likelihood of being in strict mode is higher than parsing a delete. Otherwise, LGTM

ariya:
> Super tiny micro-optimization...

It's likely to be negligible, also because `this.context.strict` is cheaper (Boolean value check), vs a string comparison of `expr.operator`.

mikesherov:
:+1:

PR #1545, opened 2016-09-06
Fix incorrect handling of duplicated proto shorthand fields

### LINKED ISSUE #1485: Duplicate '__proto__' shorthand fields should be legal
Per [B.3.1](https://tc39.github.io/ecma262/#sec-__proto__-property-names-in-object-initializers), it is only illegal to have multiple properties named `__proto__` in object literals if at least two of those properties are obtained from productions of the form `PropertyDefinition : PropertyName : AssignmentExpression`, i.e., not shorthand properties (or method definitions etc, but that case is already handled correctly). In particular, this means that `({__proto__: null, __proto__});` is a legal program.

The following tests are thus incorrect: [1](https://github.com/jquery/esprima/blob/master/test/fixtures/ES6/object-initialiser/invalid-proto-identifier-shorthand.js), [2](https://github.com/jquery/esprima/blob/master/test/fixtures/ES6/object-initialiser/invalid-proto-literal-shorthand.js), [3](https://github.com/jquery/esprima/blob/master/test/fixtures/ES6/object-initialiser/invalid-proto-shorthand-identifier.js), [4](https://github.com/jquery/esprima/blob/master/test/fixtures/ES6/object-initialiser/invalid-proto-shorthand-literal.js), [5](https://github.com/jquery/esprima/blob/master/test/fixtures/ES6/object-initialiser/invalid-proto-shorthands.js).

See also #1225.

ariya:
Fixes #1485


bakkot:
lgtm. I can't find a test that `({__proto__ = 0, __proto__ = 0} = {})` is accepted; it might be good to add that also. (I believe it is incorrectly rejected on master.)

ariya:
Good point on the test, I've added that one as well. Thanks @bakkot!

src/parser.ts
```diff
@@ -794,16 +794,6 @@ export class Parser {
             (key.type === Syntax.Literal && key.value === value);
     }
 
-    checkDuplicatedProto(key, hasProto) {
```
ariya: Also, since this is now called only from one place, the function is completely inlined.

Issue #1272, opened 2015-08-05
Customizing parsing options for the test case generator


ariya:
For a new test case, currently the baseline (`.tree.json`) is always created with the following parsing options: `{loc: true, range: true}`. This is not ideal because we want to generate the test case for e.g. tolerant parsing.

ariya:
One proposal than @jasonLaster and I were discussing is an out-of-band information for the parsing options. Thus, each test case is arranged as in this example:

  * `foo.js` is the source to be parsed
  * `foo.options.json` is the parsing options used to parse it
  * `foo.tree.json` is the expected syntax tree (baseline)

If the corresponding options file does not exist, it is assumed to be the above default.

ariya:
Side note: this is probably necessary so that we can complete #1270.

jasonLaster:
https://gist.github.com/jasonLaster/10e46802c08de1d99e9d

ariya:
Looks like we are not pursuing this at all. Will reopen if the idea is revived.

PR #1548, opened 2016-09-09
Adjust the error message for missing const initializer

### LINKED ISSUE #1547: Adjust the error message for missing const initializer
Since we tend to follow V8/Node.js, a missing initializer for `const` needs to have a similar error message.

Test case:
```js
esprima.parse('const x;')
```

Actual:
```
Error: Line 1: Unexpected token ;
```

Expected (just as V8):
```
Error: Line 1: Missing initializer in const declaration
```

ariya:
Fixes #1547

### DIFF
--- a/src/messages.ts
+++ b/src/messages.ts
@@ -46,5 +46,6 @@ export const Messages = {
     InvalidModuleSpecifier: 'Unexpected token',
     IllegalImportDeclaration: 'Unexpected token',
     IllegalExportDeclaration: 'Unexpected token',
-    DuplicateBinding: 'Duplicate binding %0'
+    DuplicateBinding: 'Duplicate binding %0',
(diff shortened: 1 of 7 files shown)


ariya:
Whoa, the downstream test failed because one of recast's test relies on the old error message. I need to investigate...

mikesherov:
:+1:

mikesherov:
As an aside, it would be nice if failure messages were JSON pretty printed like they were during the initial conversion.

ariya:
> As an aside, it would be nice if failure messages were JSON pretty printed...

At some point, we need to normalize those failure reports anyway (#1272).

ariya:
Oh, it's too bad that the workaround for recast hard-coded test worked in this pull request but it doesn't work after it's merged to master 😿

PR #1555, opened 2016-09-10
Downstream test: another attempt to workaround recast test

ariya:
Refs #1548


ariya:
This seems to work as well. Let's if it doesn't fail once it's merged.

PR #1559, opened 2016-09-12
Adjust the error message for disallowed bound name

### LINKED ISSUE #1558: Adjust the error message for disallowed bound name
Since we tend to follow V8/Node.js, an invalid bound name needs to have a similar error message.

Test case:
```js
esprima.parse('let let');
```

Actual:
```
Error: Line 1: Unexpected token let
```

Expected (just as V8):
```
Error: Line 1: let is disallowed as a lexically bound name
```

ariya:
Fixes #1558


mikesherov:
:+1:

PR #1560, opened 2016-09-12
Do not implicitly collect comments when comment attachment is specified

### LINKED ISSUE #1553: Comment attachment should not imply comment collection
Test case:
```js
esprima.parse('x // some info', { attachComment: true, comment: false }).comments
```
Actual:
```
[ { type: 'Line', value: ' some info' } ]
```
Expected:
```
undefined
```

Comment collection that results in the `comments` array, **must** happen only when `comment: true` is specified in the parsing option. It **must not** be _implicitly_ enabled when `attachComment: true` is specified.

ariya:
Fixes #1553

### DIFF
--- a/src/esprima.ts
+++ b/src/esprima.ts
@@ -39,8 +39,9 @@ export function parse(code, options, delegate) {
     };
 
     let parserDelegate = (typeof delegate === 'function') ? proxyDelegate : null;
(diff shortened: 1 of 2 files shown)


mikesherov:
:+1:

PR #1565, opened 2016-09-18
Downstream test: include Istanbul again

ariya:


ariya:
Still not working, very unfortunate 😿

PR #1567, opened 2016-09-19
Support ES2017 trailing commas in function parameter lists and calls

### LINKED ISSUE #1550: ES2017: Trailing commas in function parameter lists and calls
The [trailing commas in function parameter lists and calls](https://jeffmo.github.io/es-trailing-function-commas/) proposal is now Stage 4 and included in the [finished proposals](https://github.com/tc39/proposals/blob/master/finished-proposals.md) list.

This was brought up a year ago in Issue #316 with a Pull Request at the time https://github.com/facebookarchive/esprima/pull/104/commits/5650ab0221b6bb569e95a940e8eac76a6d48d654 but was deferred since the proposal was only Stage 1. Now that it has moved forward it would be worth re-investigating. Some JavaScript engines already support it in technology previews.

ariya:
Fixes #1550


mikesherov:
Seems to be missing a test for valid rest? `function f(...a,) {}`

ariya:
> Seems to be missing a test for valid rest? function f(...a,) {}

Good catch, I'll add it (and it seems to be a syntax error).

mikesherov:
:+1:

PR #1571, opened 2016-09-28
Improve class declaration parsing for the default export

### LINKED ISSUE #1570: Parsing exported class declaration should not internally mutate the node
When the following construct is being parsed:
```javascript
export default class {};
export default class A {};
```
internally each export declaration has its `class` parsed as a ClassExpression and _then_ the node type is changed to ClassDeclaration.

For clarity and consistency with the exported function declaration, the above approach should rewritten so that internal node mutation does not happen.

ariya:
When parsing `export default class`, treat it as a real class declaration
and do not perform the two-step process (parse as a class expression and then
mutate the node into a class declaration).

Fixes #1570

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -1,5 +1,5 @@
 import { assert } from './assert';
-import { Messages} from './messages';
+import { Messages } from './messages';
 
 import { ErrorHandler } from './error-handler';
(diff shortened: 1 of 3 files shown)


mikesherov:
:+1:

PR #1572, opened 2016-09-29
Use TypeScript 2.0

### LINKED ISSUE #1569: Use TypeScript 2.0
Among others, this allows us to leverage strict null checks.

References:
* [TypeScript 2.0 announcement](https://blogs.msdn.microsoft.com/typescript/2016/09/22/announcing-typescript-2-0/)
* [What's new in TypeScript](https://github.com/Microsoft/TypeScript/wiki/What%27s-new-in-TypeScript)

ariya:
Various fixes necessary to satistify strict null checking:

* Safeguard against possible null value
* Initialize values before they are used
* Annotate arrays because there is no type widening anymore
* Inline throw so that the control flow is determined correctly

Other TypeScript 2.0 features being utilized:

* Use `readonly` property feature whenever appropriate
* Remove unused local variables
* Remove unused function parameters
* Unify return paths

Also, bump TypeScript formatter to the version that uses TypeScript 2.

Fixes #1569

### DIFF
--- a/src/error-handler.ts
+++ b/src/error-handler.ts
@@ -9,7 +9,7 @@ declare class Error {
 }
 
 export class ErrorHandler {
-    errors: Error[];
+    readonly errors: Error[];
     tolerant: boolean;
 
     constructor() {
@@ -42,6 +42,8 @@ export class ErrorHandler {
         } finally {
             return error;
         }
+        /* istanbul ignore next */
+        return error;
     };
 
     createError(index: number, line: number, col: number, description: string): Error {
--- a/src/nodes.ts
+++ b/src/nodes.ts
@@ -1,11 +1,12 @@
 import { Syntax } from './syntax';
 
 export type ArgumentListElement = Expression | SpreadElement;
-export type ArrayExpressionElement = Expression | SpreadElement;
-export type ArrayPatternElement = AssignmentPattern | BindingIdentifier | BindingPattern | RestElement;
+export type ArrayExpressionElement = Expression | SpreadElement | null;
+export type ArrayPatternElement = AssignmentPattern | BindingIdentifier | BindingPattern | RestElement | null;
 export type BindingPattern = ArrayPattern | ObjectPattern;
 export type BindingIdentifier = Identifier;
 export type Declaration = ClassDeclaration | ExportDeclaration | FunctionDeclaration | ImportDeclaration | VariableDeclaration;
+export type ExportableNamedDeclaration = ClassDeclaration | Function | VariableDeclaration;
 export type ExportDeclaration = ExportAllDeclaration | ExportDefaultDeclaration | ExportNamedDeclaration;
 export type Expression = ArrayExpression | ArrowFunctionExpression | AssignmentExpression |
     BinaryExpression | CallExpression | ClassExpression | ComputedMemberExpression |
@@ -23,8 +24,8 @@ export type PropertyValue = AssignmentPattern | BindingIdentifier | BindingPatte
 export type StatementListItem = Declaration | Statement;
 
 export class ArrayExpression {
-    type: string;
-    elements: ArrayExpressionElement[];
+    readonly type: string;
+    readonly elements: ArrayExpressionElement[];
     constructor(elements: ArrayExpressionElement[]) {
         this.type = Syntax.ArrayExpression;
         this.elements = elements;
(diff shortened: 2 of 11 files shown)


src/nodes.ts
```diff
@@ -22,31 +23,33 @@ export type PropertyKey = Identifier | Literal;
 export type PropertyValue = AssignmentPattern | BindingIdentifier | BindingPattern | FunctionExpression;
 export type StatementListItem = Declaration | Statement;
 
+type Nullable<T> = T | null;
```
michaelficarra: I don't like this `Nullable<T>` type. It's deceptive. You might think it's something similar to a `Maybe`/`Option`, but it's not. `Nullable<Nullable<T>>` has T+1 inhabitants while `Maybe<Maybe<T>>` has T+2. This is a very important distinction.

ariya: What do you suggest instead? How about `Maybe<T>`?

michaelficarra: It occurs to me that you're actually *looking for* the `T | null` type. In that case, I say you just use ` | null` wherever you want this.

ariya: That was my original implementation. There were however 45 occurrences of `| null` and hence I looked for a sensible shorthand for that.

ariya: I was also toying with the idea of e.g. `type MaybeIdentifier = Identifier | null` and use `MaybeIdentifier` at the right places. It didn't feel much worse or much better.

michaelficarra: I don't think there's anything wrong with just `| null` all over the place. It's immediately obvious exactly what the type is.

ariya: I don't have any strong preference on one or another, I'll revert it back to use `| null` again. Beside, I have the feeling that TypeScript will have `Nullable<T>` or `?T` at some point (which is proven to be quite popular in C# already).

mikesherov:
I defer to @michaelficarra on review here.

michaelficarra:
Oh please don't. I don't know anything about TypeScript.

mikesherov:
you know enough to say: `I don't like this Nullable<T> type. It's deceptive. You might think it's something similar to a Maybe/Option, but it's not. Nullable<Nullable<T>> has T+1 inhabitants while Maybe<Maybe<T>> has T+2. This is a very important distinction.`

michaelficarra:
lol fair. But I'm not qualified to review this PR, really.

src/error-handler.ts
```diff
@@ -42,6 +42,8 @@ export class ErrorHandler {
         } finally {
             return error;
         }
+        /* istanbul ignore next */
+        return error;
```
DanielRosenwasser: @ariya Any reason you need this?

src/parser.ts
```diff
@@ -669,13 +669,13 @@ export class Parser {
                     } else if (this.matchKeyword('class')) {
                         expr = this.parseClassExpression();
                     } else {
-                        this.throwUnexpectedToken(this.nextToken());
+                        throw this.unexpectedTokenError(this.nextToken());
```
DanielRosenwasser: This is fine but if you were intent on using the same method as before for whatever reason, you could make throwUnexpectedToken return never and then return the result of that.

src/error-handler.ts
ariya: Without that, I got this error:
```
error TS2366: Function lacks ending return statement and return type does not include 'undefined'
```

src/parser.ts
ariya: Whoa, thanks for that `never` trick!
I'll revert those inlined throws and just use your suggested approach.

src/error-handler.ts
DanielRosenwasser: I think that might be a bug on us. Glad I checked in!

src/nodes.ts
johanssj: Since I don't like disguising types that union in null (because they get lost in the maze with types that don't admit null & you don't want to have to remember which types do and which types don't), I have started using the following naming convention in my own TS2 projects.

`type IdentifierQ = Identifier | null`

The intent is that the letter 'Q' stands for 'question mark' as in the Kleene operator '?' or zero-or-one occurrences of values of the stem type.

Also I don't mind 'S' as another identifier suffix which stands for 'star' as in Kleen '*'. This is useful for array types & other collections which may contain zero-or-more occurrences of said value.

Personally I find this to be a low-noise naming convention; of course everyone has there own stylistic preferences.

Acknowledgments to XPath and XQuery type syntax for the inspiration (also occurrence characters used in RegEx expressions).

Any suggestions for '+', the one-or-more variation?

PR #1577, opened 2016-10-03
Explicitly distinguish parsing a module vs a script

### LINKED ISSUE #1576: [API] Explicitly distinguish parsing a module vs a script
Since ES2015, a program can be a module or a script. To correctly parse the source input, the `sourceType` option needs to be specified (when omitted, it is assumed to be `script`). This is illustrated in the following example:

```javascript
// parsing a script
esprima.parse('x = 2', { sourceType: 'script' });

// parsing a module
esprima.parse('export const y = 3', { sourceType: 'module' });
```

To encourage the consumer to identify the source input, two new functions shall be introduced, `parseScript` and `parseModule`. The previous code fragment is then equivalent to the following:
```javascript
// parsing a script
esprima.parseScript('x = 2');

// parsing a module
esprima.parseModule('export const y = 3');
```

At the same time, `parse` should be deprecated. It will be still available in the near future for backward compatibility but the use of it should be highly discouraged.

References:
* Shift Parser: http://shift-ast.org/parser.html
* Common misunderstanding: Esprima doesn't understand export/import #1273 #1336
* Original pull request for `sourceType`: #290

--

ariya:
`parse` is still supported but the use of it is discouraged.
The consumer must choose between `parseScript` and `parseModule` to perform
a syntactical analysis on a script or a module, respectively.

Fixes #1576

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -18,6 +18,7 @@ interface Config {
 }
 
 interface Context {
+    isModule: boolean;
     allowIn: boolean;
     allowYield: boolean;
     firstCoverInitializedNameError: any;
@@ -60,7 +61,6 @@ export class Parser {
     readonly scanner: Scanner;
     readonly operatorPrecedence: any;
 
-    readonly sourceType: string;
     lookahead: any;
     hasLineTerminator: boolean;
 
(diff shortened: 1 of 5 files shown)


src/parser.ts
```diff
@@ -60,7 +61,6 @@ export class Parser {
     readonly scanner: Scanner;
     readonly operatorPrecedence: any;
 
-    readonly sourceType: string;
```
mikesherov: What's the reason for changing this over to `isModule`? Previously you had argued for a string type to support eventual additional sourceTypes. Is that YAGNI now?

ariya: That's right. This variable is internal to the parser since the only use of `sourceType` was to compare it against `module`.

`sourceType` as a visible property of `Program` node still exists and hasn't changed.

PR #1580, opened 2016-10-04
Fix incorrect parsing of export specifier

### LINKED ISSUE #1578: IdentifierName should be used for ExportSpecifier
Test case:
```javascript
esprima.parse('export { if } from "abc"', { sourceType: 'module' });
```

**Actual**: An exception is thrown:
```
Error: Line 1: Unexpected token if.
```

**Expected**: A valid syntax tree:
```
Program {
  type: 'Program',
  body: 
   [ ExportNamedDeclaration {
       type: 'ExportNamedDeclaration',
       declaration: null,
       specifiers: [Object],
       source: [Object] } ],
  sourceType: 'module' }
```

Per ES2015 grammar, the production for ExportSpecifier:

```
ExportSpecifier :
  IdentifierName
  IdentifierName as IdentifierName
```

ariya:
Instead of `Identifier`, use `IdentifierName` in `ExportSpecifier`.

Fixes #1578

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -3223,7 +3223,7 @@ export class Parser {
     parseExportSpecifier(): Node.ExportSpecifier {
         const node = this.createNode();
 
-        const local = this.matchKeyword('default') ? this.parseNonComputedProperty() : this.parseVariableIdentifier();
+        const local = this.parseNonComputedProperty();
         let exported = local;
         if (this.matchContextualKeyword('as')) {
             this.nextToken();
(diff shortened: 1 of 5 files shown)


src/parser.ts
```diff
@@ -3223,7 +3223,7 @@ export class Parser {
     parseExportSpecifier(): Node.ExportSpecifier {
         const node = this.createNode();
 
-        const local = this.matchKeyword('default') ? this.parseNonComputedProperty() : this.parseVariableIdentifier();
+        const local = this.parseNonComputedProperty();
```
ariya: The name is still kind of confusing. Expect a future refactoring that changes it to `parseIdentifierName`.

PR #1583, opened 2016-10-07
Update references to ECMA-262 7th edition

ariya:


src/parser.ts
```diff
@@ -1334,13 +1334,14 @@ export class Parser {
...
     }
 
-    // ECMA-262 12.6 Multiplicative Operators
-    // ECMA-262 12.7 Additive Operators
-    // ECMA-262 12.8 Bitwise Shift Operators
-    // ECMA-262 12.9 Relational Operators
-    // ECMA-262 12.10 Equality Operators
-    // ECMA-262 12.11 Binary Bitwise Operators
-    // ECMA-262 12.12 Binary Logical Operators
+    // ECMA-262 12.6 Exponentiation Operators
```
ariya: We need to seriously consider that for the later editions. It's still problematic either way for ES2017, the chapter index is still not known (for e.g. stage-4 async function) and the URL doesn't exist yet (http://ecma-international.org/ecma-262/8.0), until the spec is almost finalized.

src/parser.ts
```diff
@@ -2915,7 +2916,7 @@ export class Parser {
         return this.finalize(node, new Node.FunctionExpression(null, params.params, method, isGenerator));
     }
 
-    // ECMA-262 14.4 Yield expression
+    // ECMA-262 14.4 Generator Function Definitions
```
michaelficarra: This doesn't seem right.

mathiasbynens:
@michaelficarra Usually I’d agree, but in this case I see why @ariya prefers linking to a versioned snapshot. Linking to the living spec could cause a mismatch between what Esprima implements and what the spec says.

ariya:
@michaelficarra Discuss it here: #1585.

PR #1594, opened 2016-10-15
Test fixture: simplify module parsing

### LINKED ISSUE #1593: Test fixtures: simplify module parsing
Currently, the set of test fixtures (under `test/fixtures`) is not easy to be expanded with additional test cases for module parsing (i.e. `sourceType` of `module`).

The proposed step to fix it is rather easy: indicate `sourceType` by the file name, e.g. `foo.module.js` is treated as a module (instead of a script). This is both used for creating a new test fixture (if the baseline does not exist yet) or to evaluate an existing test case.

ariya:
For every test case, the source type is indicated by the file name's suffix:
.module.js indicates a module, otherwise it is a script.

Fixes #1593


mikesherov:
:+1:

Issue #1079, opened 2015-02-25
ES2017: Async Function and await expression


ariya:
New productions for async functions:

```
AsyncFunctionDeclaration[Yield, Await, Default] ::
    async [no LineTerminator here] function BindingIdentifier[?Yield, ?Await] ( FormalParameters[Await] ) { AsyncFunctionBody }
    [+Default] async [no LineTerminator here] function ( FormalParameters[Await] ) { AsyncFunctionBody }

AsyncFunctionExpression ::
    async [no LineTerminator here] function ( FormalParameters[Await] ) { AsyncFunctionBody }
    async [no LineTerminator here] function BindingIdentifier[Await] ( FormalParameters[Await] ) { AsyncFunctionBody }

AsyncMethod[Yield, Await] ::
    async [no LineTerminator here] PropertyName[?Yield, ?Await] ( StrictFormalParameters[Await] ) { AsyncFunctionBody }

AsyncFunctionBody ::
    FunctionBody[Await]

AsyncArrowFunction[In, Yield, Await] ::
    async [no LineTerminator here] AsyncArrowBindingIdentifier[?Yield] [no LineTerminator here] => AsyncConciseBody[?In]
    CoverCallExpressionAndAsyncArrowHead[?Yield, ?Await] [no LineTerminator here] => AsyncConciseBody[?In]

AsyncConciseBody[In] ::
    [lookahead != {] AssignmentExpression[?In, Await]
    { AsyncFunctionBody }

AsyncArrowBindingIdentifier[Yield]
    BindingIdentifier[?Yield, Await]

CoverCallExpressionAndAsyncArrowHead[Yield, Await] ::
    MemberExpression[?Yield, ?Await] Arguments[?Yield, ?Await]

AsyncArrowHead ::
    async [no LineTerminator here] ArrowFormalParameters[Await]
```

where they are being used in various places:

```
HoistableDeclaration[Yield, Default, Await] ::
    AsyncFunctionDeclaration[?Yield, ?Default, ?Await]

PrimaryExpression[Yield, Await] ::
    AsyncFunctionExpression

MethodDefinition[Yield, Await] ::
    AsyncMethod[?Yield, ?Await]

AssignmentExpression[In, Yield, Await] ::
    AsyncArrowFunction[?In, ?Yield, ?Await]

CallExpression[Yield, Await] ::
    CoverCallExpressionAndAsyncArrowHead[?Yield, ?Await]
```

New production for await expression:

```
[...]

PR #1595, opened 2016-10-20
Introduce `async` property to function nodes

ariya:
Refs #1079


ariya:
This is intentionally a separate check-in so that the forthcoming async function implementation will not touch hundreds of fixture baselines.

PR #1598, opened 2016-10-24
add a compile-es2015 npm script

toshok:
https://github.com/toshok/echojs uses esprima and at the moment uses an ancient branch where the IIFE has been manually replaced by an es2015 module.

In the interest of tracking closer to master, this change adds a `compile-es2015` script which generates `dist/es2015/` containing all the files as es2015 modules.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ node_modules
 test/dist
 src/*.js
 dist/esprima.js
+dist/es2015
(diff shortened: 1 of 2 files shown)


ariya:
Could you open an issue to discuss this? It triggers a bigger decision on the best way to distribute the library (ES5 vs ES.next). I'm not fancy doing it without a proper plan, e.g. how to test it.

toshok:
done (#1599).

ariya:
Let's hold this off while we're figuring out the test strategy (#1599).

PR #1607, opened 2016-10-30
Detect hex digits correctly

### LINKED ISSUE #1605: It's missing "Unexpected token ILLEGAL" exception for 'g', 'G', 'h' and 'H' tokens inside hexadecimal digits
```javascript
"\u00FG" // No error
```

```javascript
"\xFG" // No error
```

```javascript
0xFG // No error
```

My browser currently throws an "Invalid Unicode escape sequence" / "Invalid or unexpected token" exception for each of these tests above, while Esprima considers it valid. Tested at http://esprima.org/demo/parse.html.

ariya:
Fixes #1605


ariya:
The failing runs from Travis CI are false alarm. For some strange reasons, TypeScript Formatter complains that the source files are not formatted properly.

PR #1616, opened 2016-11-04
Ensure that import specifier is more restrictive

### LINKED ISSUE #1615: Import specifier should be more restrictive
Test case:
```
esprima.parse('import { null } from "null"', { sourceType: 'module' })
```

**Expected**: A thrown exception.

**Actual**: A valid syntax tree:
```js
Program {
  type: 'Program',
  body: 
   [ ImportDeclaration {
       type: 'ImportDeclaration',
       specifiers: [Object],
       source: [Object] } ],
  sourceType: 'module' }
```

**Note**: The example source snippet is invalid because there are only two valid productions for `ImportSpecifier`:
```
ImportSpecifier :
    ImportedBinding
    IdentifierName as ImportedBinding
```
A valid example of the second production will be e.g. `import { null as nil } from "null"`.

ariya:
Fixes #1615

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -3274,13 +3274,24 @@ export class Parser {
     parseImportSpecifier(): Node.ImportSpecifier {
         const node = this.createNode();
 
-        let local;
-        const imported = this.parseIdentifierName();
-        if (this.matchContextualKeyword('as')) {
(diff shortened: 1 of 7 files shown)


mikesherov:
👍

PR #1621, opened 2016-11-06
Update copyright notice

ariya:
/cc @kborchers


kborchers:
This may be easier and completely up to you but we don't have any requirement for you to have license/copyright headers in every file so you could just delete those and make sure the LICENSE file is updated.

ariya:
@kborchers Let's do that separately so that this step (copyright update) is memorialized first for the benefit of any future `git log` or `git blame`.

PR #1631, opened 2016-11-21
Improve XHTML entity scanning in JSX

### LINKED ISSUE #1629: XHTML entities handling (in JSX) needs to be more robust
Test case:
```js
esprima.parse('<a b="&#123"/>', { jsx: true });
```
**Expected**: A syntax tree.

**Actual**: An exception `Unexpected token ILLEGAL`.

**Note**: In this case, `&#123` is not a valid XHTML entity (it does not terminate with `;`). Hence, it should be treated as a regular character sequence and it should not be tokenized as an XHTML entity. This is consistent with the behavior of other implementations such as Flow and TypeScript.

Other similar cases:

  * Incomplete: `&;`
  * Non-numeric: `&#1z;`
  * Non-hex: `&#xAZ;`
  * Uppercase: `&#X41;`

ariya:
Fixes #1629


mikesherov:
:+1:

PR #1648, opened 2016-12-01
Update scanner.ts

hydroper:
Added basic error handling for keywords with escaped characters

### DIFF
--- a/src/scanner.ts
+++ b/src/scanner.ts
@@ -392,13 +392,18 @@ export class Scanner {
             }
         }
 
-        return this.source.slice(start, this.index);
+        return {
+            escapedChars: 0,
+            id: this.source.slice(start, this.index);
+        };
     };
 
     getComplexIdentifier(): string {
         let cp = this.codePointAt(this.index);
         let id = Character.fromCodePoint(cp);
         this.index += id.length;
+
+        let escapedChars = 0;
 
         // '\u' (U+005C, U+0075) denotes an escaped character.
         let ch;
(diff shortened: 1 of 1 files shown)


ariya:
Thanks @hydroper! Please check our CONTRIBUTING.md. Tweak the commit message to link to the issue and to have more descriptive first line (use `git log` to see past commit logs as examples). Please add some tests, `npm test` should not fail.

src/scanner.ts
```diff
@@ -498,6 +510,10 @@ export class Scanner {
             type = Token.Identifier;
         }
 
+        if (idResult.escapedChars && type !== Token.Identifier)
+            this.errorHandler.throwError(this.index, this.lineNumber,
```
ariya: Use `throwUnexpectedToken` instead.

src/scanner.ts
```diff
@@ -392,14 +392,19 @@ export class Scanner {
             }
         }
 
-        return this.source.slice(start, this.index);
+        return {
+            escapedChars: 0,
```
ariya: I'm afraid this approach will regress scanning performance, due to a object created for every single tokenized identifier.

src/scanner.ts
```diff
@@ -392,14 +392,19 @@ export class Scanner {
...
+            id: this.source.slice(start, this.index);
```
ariya: Semicolon here won't work.

src/scanner.ts
hydroper: Oh, true. In this case it'd be possible to anotate the number of escaped characters in some property at the scanner class, or a variable.

hydroper:
Sorry, I'm quite new here yet ! I'll try to follow this CONTRIBUTING.md

PR #1650, opened 2016-12-02
ES2017: Handle non-identifier async method name for class element

### LINKED ISSUE #1649: Async method with non-identifier name as a class element
This is similar to #1626 but applying to class instead of object.

Test input:
```
class X { async "f"(){} }
class X { async 1(){} }
class X { async ["f"](){} }
```

Expected: each line produces a valid AST.
Actual: each line triggers an error.

ariya:
Fix #1649


mikesherov:
👍

PR #1655, opened 2016-12-04
Upgrade to TSLint 4

ariya:
`tslint.json` is massively simplified since most rules are already covered by
tslint:recommended.


mikesherov:
:+1:

PR #1658, opened 2016-12-07
Remove incorrect test (already covered by `update_exp.js`)

### LINKED ISSUE #1636: invalid update is not invalid
@ariya I took a deep dive into the tests, and found various strange things. One of the strange things discovered is this test: https://github.com/jquery/esprima/blob/master/test/fixtures/ES2016/exponent/invalid_update.js

Run this test in the online parser: http://esprima.org/demo/parse.html

It does not throw, but got parsed!

I also ran the check with latest Esprima version, and got same result. However. I also found some other similar cases. So there are many tests that throws that shouldn't throw. And many tests that shouldn't throw, that throws.

Another similar case is this to prove my stand: https://github.com/jquery/esprima/blob/master/test/fixtures/ES2016/exponent/invalid_typeof_exp.js

ariya:
Fixes #1636


mikesherov:
:+1:

PR #1659, opened 2016-12-07
Limit function declaration as if statement clause only in non-strict mode

### LINKED ISSUE #1657: Function declaration as if statement clause should be prohibited in strict mode
Test case:
```js
esprima.parse('"use strict"; if (true) function f(){}');
```
Actual output: a syntax tree.

Expected output: an exception indicating a syntax error.

<hr>

For the syntax error, the message ([following V8](https://chromium.googlesource.com/v8/v8/+/master/src/messages.h)) should be:

> In strict mode code, functions can only be declared at top level or inside a block

<hr>

Section B.3.4 on _FunctionDeclarations in IfStatement Statement Clauses_ includes the following productions:

```
IfStatement :
    if ( Expression ) FunctionDeclaration else Statement
    if ( Expression ) Statement else FunctionDeclaration
    if ( Expression ) FunctionDeclaration else FunctionDeclaration
    if ( Expression ) FunctionDeclaration
```

And then states that:

> The above rules are only applied when parsing code that is not strict mode code.

Non-strict mode is permitted to follow the above productions due to web compatibility reasons (see also #1051).

ariya:
Fixes #1657

### DIFF
--- a/src/messages.ts
+++ b/src/messages.ts
@@ -33,6 +33,7 @@ export const Messages = {
     StrictLHSPostfix: 'Postfix increment/decrement may not have eval or arguments operand in strict mode',
     StrictLHSPrefix: 'Prefix increment/decrement may not have eval or arguments operand in strict mode',
     StrictReservedWord: 'Use of future reserved word in strict mode',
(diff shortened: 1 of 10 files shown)


mikesherov:
👍

PR #1663, opened 2016-12-09
Account for different possible arguments of a yield expression

### LINKED ISSUE #1469: a ? yield : b fails to parse
Try this code on http://esprima.org/demo/rewrite.html

``` javascript
function* gf() {
    var foo = false;
    return foo ? (yield* (yield)) : 10;
}
```

The first rewrite removes the parentheses around the `yield*` and `yield` expressions.  This produces syntactically invalid code which you can see by then pressing the rewrite button again, and esprima complains that an unexpected colon token was found.

ariya:
Fix #1469

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -3090,6 +3090,33 @@ export class Parser {
 
     // ECMA-262 14.4 Generator Function Definitions
 
+    isStartOfExpression(): boolean {
+        let start = true;
+
+        const value = this.lookahead.value;
+        switch (this.lookahead.type) {
+            case Token.Punctuator:
+                start = (value === '[') || (value === '(') || (value === '{') ||
+                    (value === '+') || (value === '-') ||
+                    (value === '!') || (value === '~') ||
+                    (value === '++') || (value === '--') ||
+                    (value === '/') || (value === '/=');  // regular expression literal
+                break;
+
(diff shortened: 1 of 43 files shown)


mikesherov:
Looks good except it's missing a test for `super`.

mikesherov:
Also, not sure if we need a `yield yield` test?

ariya:
> Also, not sure if we need a yield yield test?

For that, there exists already a test.

ariya:
> Looks good except it's missing a test for super.

I'm still not sure what would be a good test for this. I speculate that we'll need it only once `isStartOfExpression` is being used for other purposes (than just parsing yield's argument).

mikesherov:
Forgive me. Is this not a valid test of yield super?

```
class A extends B {
  *x() {
    yield super();
  }
}
```

ariya:
@mikesherov Unfortunately not: `super` needs to be in a constructor (which can't be a generator).

mikesherov:
@ariya here's a jsfiddle: https://jsfiddle.net/vwsrL6yL/

ariya:
@mikesherov That looks like a good test, I'll add it. Thanks!

mikesherov:
Once the test is added... :+1:

PR #1665, opened 2016-12-10
Prohibit labelled class/generator/function declaration

### LINKED ISSUE #1484: Incorrectly accepts invalid LabeledStatement (ES2015)
If I understand the spec correctly ([§13.13](http://www.ecma-international.org/ecma-262/6.0/index.html#sec-labelled-statements)) , the following statements should result in a syntax errors:

```
a: class abc {}
c: function* bar() {}
```

Grammar:

```
LabelledStatement:
    LabelIdentifier : LabelledItem

LabelledItem :
  Statement
  FunctionDeclaration
```

`let`/`const` declarations, generation functions and class declarations are neither `Statement`s nor `FunctionDeclaration`s.

FWIW, most other parser seem to be behave the same way. Only shift and traceur are throwing errors. This makes me wonder whether this is an intentional deviation from the spec for compatibility reasons? I also opened an issue about this for acorn: https://github.com/ternjs/acorn/issues/393 .

http://astexplorer.net/#/QOCdiNhTai (the second example is commented out because esprima throws the error "Line 6: Use of future reserved word in strict mode", so it seems it got that one right at least?)

ariya:
The only allowed situation is labelled function declaration in non-strict mode
(for web compatibility reason).

Fix #1484


ariya:
Please ignore the failing downstream test. It's is a false alarm, it's due to Assetgraph not working with the latest ESLint (see https://github.com/assetgraph/assetgraph/pull/662 for more details).

mikesherov:
Super interesting set of tests. I learn more about JS reviewing these PRs then anywhere else. :+1:

PR #1666, opened 2016-12-11
Downstream test: temporarily workaround AssetGraph linting issue

ariya:
Refs #1665


ariya:
It's working! Looks like we ought to use this workaround to prevent incomplete downstream tests.

mikesherov:
Sneaky!

PR #1669, opened 2016-12-13
Only permit hex digits in hex escape sequence

### LINKED ISSUE #1619: Incorrectly accepts \x{0} in string/template literals
```js
'\x{0}'
```

is not a program; nor is

```js
`\x{0}`
```

----

Fix is to move [this if clause](https://github.com/jquery/esprima/blob/d044f5b5b708451fb24405493777997be887f46e/src/scanner.ts#L824-L827) and [this if clause](https://github.com/jquery/esprima/blob/d044f5b5b708451fb24405493777997be887f46e/src/scanner.ts#L943-L946) into the respective `'u'` cases, I believe.

ariya:
Adjust the error message to follow V8.
Also make it explicit that scanHexEscape can return null (upon an
invalid sequence).

Fixes #1619


mikesherov:
:+1:

PR #1671, opened 2016-12-14
Prohibit any escape sequence in a reserved word

### LINKED ISSUE #1612: No exception for keywords with escaped characters
**Test input**

```javascript
\u{74}rue
```

**Expected:** an error.

**Actual:** produces a valid AST.

ariya:
Fix #1612


hydroper:
I think I got it... the tests are scripts stored in the test directory??

PR #1674, opened 2016-12-14
Reuse existed `parseSuper` method.

hzlmn:
Reuse existed `parseSuper` method for parsing LHS-expressions with allow call.

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -1238,17 +1238,8 @@ export class Parser {
         const previousAllowIn = this.context.allowIn;
         this.context.allowIn = true;
 
-        let expr;
-        if (this.matchKeyword('super') && this.context.inFunctionBody) {
-            expr = this.createNode();
-            this.nextToken();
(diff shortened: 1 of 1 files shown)


src/parser.ts
```diff
@@ -1292,14 +1283,12 @@ export class Parser {
         return expr;
     }
 
-    parseSuper(): Node.Super {
+    parseSuper(allowCall = false): Node.Super {
```
ariya: That's a design pattern we'd like to avoid. Search for "boolean trap" for more details.

hzlmn: Thanks for pointing. I read your post, sorry for that :) Any other remarks?

ariya:
If this is part of your attempt to fix `super` in constructor, I'd say combine it with the fix. There is no urgent need for a separate refactoring (also, see my comment above).

hzlmn:
Ok, got it. Yes, I discovered it while working on that part. Closing it.

PR #1680, opened 2016-12-18
Limit strict mode directive to functions with a simple parameter list

### LINKED ISSUE #1677: [ES2016] Only functions with simple parameter list can have strict mode directive
As mentioned in the specification [14.1 Function Definitions](https://tc39.github.io/ecma262/#sec-function-definitions-static-semantics-early-errors):

> It is a Syntax Error if ContainsUseStrict of _FunctionBody_ is **true** and IsSimpleParameterList of _FormalParameters_ is **false**.

The simplest test case:
```js
esprima.parse('function f(x=1) {"use strict";}');
```

Expected: An exception indicating the syntax error.

Actual: A valid syntax tree.

<hr>

Since this is affecting _FunctionBody_, the test cases shall also cover function expressions and declarations, method definitions, as well as arrow function expressions.

<hr>

Following V8, [the error message](https://chromium.googlesource.com/v8/v8/+/master/src/messages.h) for the case should be:

> SyntaxError: Illegal 'use strict' directive in function with non-simple parameter list

ariya:
Fix #1677


mikesherov:
:+1:

Mike Sherov

PR #1682, opened 2016-12-20
Rest element and rest parameter should include binding patterns

### LINKED ISSUE #1681: Rest element and rest parameter should include binding patterns
Test case for rest element:
```js
esprima.parse('let [...{a}] = b')
```

Test case for rest parameter:
```js
esprima.parse('function f(...{a}){}')
```

Expected result (for both cases): A valid syntax tree.

Actual result (for both cases): A thrown exception `Unexpected token {`.

ariya:
Fix #1681


test/fixtures/ES6/arrow-function/complex-rest-in-arrow-not-allowed.failure.json
```diff
@@ -1 +1 @@
-{"index":6,"lineNumber":1,"column":7,"message":"Error: Line 1: Unexpected token [","description":"Unexpected token ["}
\ No newline at end of file
+{"index":10,"lineNumber":1,"column":11,"message":"Error: Line 1: Strict mode function may not have duplicate parameter names","description":"Strict mode function may not have duplicate parameter names"}
```
mikesherov: Seems this test should be renamed to `complex-rest-in-arrow-with-duplicated-names-not-allowed.js` and a new test for complex rest should be created, showing it's allowed in arrows: `(a, ...[b]) => c`

ariya: Good point! I'll make those changes.

PR #1686, opened 2016-12-22
Adjust the error message for bad getter/setter arity

### LINKED ISSUE #1685: Adjust the error message for incorrect getter/setter arity
This is to align the description to follow [V8 error messages](https://chromium.googlesource.com/v8/v8/+/master/src/messages.h).

Test case for **setter method**:
```js
esprima.parse('x = { set y(a,b) { } }')
```

* Expected error message: "Setter must have exactly one formal parameter".
* Actual error message: "Unexpected token ,".

Test case for **getter method**:
```js
esprima.parse('x = { get y(a) { } }')
```

* Expected error message: "Getter must not have any formal parameters".
* Actual error message: "Unexpected identifier".

ariya:
Fix #1685


hydroper:
@ariya Can you update the esprima parser used in the online demos?

mikesherov:
@hydroper, the version on the demos gets updated when releases get cut.

hydroper:
@mikesherov Unfortunately, it looks like it's not updated here: http://esprima.org/demo/parse.html?code=%5Cu%7B66%7Dor%20(%3B%3B)%3B

It should throw an error.

ariya:
@hydroper That's not related to this PR (please don't do thread hijacking). The demo is only for stable releases, not for any development version (from master). It's updated once there is a 4.0 release.

hydroper:
@ariya I could have issued that. I got it anyways.

PR #1691, opened 2016-12-23
README: Add a link to the complete documentation

### LINKED ISSUE #1687: Explicitly link to the complete documentation from the README
Currently the README [says](https://github.com/jquery/esprima/blob/8391d896a5c260191d7885325302a339f3b4e026/README.md#api):

> ### API
> 
> Esprima can be used to perform [lexical analysis](https://en.wikipedia.org/wiki/Lexical_analysis) (tokenization) or [syntactic analysis](https://en.wikipedia.org/wiki/Parsing) (parsing) of a JavaScript program.
> 
> A simple example on Node.js REPL:
> 
> ```javascript
> > var esprima = require('esprima');
> > var program = 'const answer = 42';
> 
> > esprima.tokenize(program);
> [ { type: 'Keyword', value: 'const' },
>   { type: 'Identifier', value: 'answer' },
>   { type: 'Punctuator', value: '=' },
>   { type: 'Numeric', value: '42' } ]
>   
> > esprima.parse(program);
> { type: 'Program',
>   body:
>    [ { type: 'VariableDeclaration',
>        declarations: [Object],
>        kind: 'const' } ],
>   sourceType: 'script' }
> ```

And this is confusing, it made me think that it's the entire API, nothing else. But there are many options which are missing as well as not immediately discoverable from the README (for example, `sourceType`, which changes many things and omitting it results in weird errors).

Please at least adding a link to this page: http://esprima.org/doc/index.html saying that "see complete documentation here".

ariya:
Fix #1687

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,3 +42,5 @@ A simple example on Node.js REPL:
        kind: 'const' } ],
   sourceType: 'script' }
 ```
+
+For more information, please read the [complete documentation](http://esprima.org/doc).


mikesherov:
:+1:

Mike Sherov

PR #1704, opened 2016-12-31
Do not mutate the token object

### LINKED ISSUE #1702: Eliminate token object mutation
In some cases, the token object is still mutated (properties being added or modified), see e.g. `scanPunctuator` and `parseBinaryExpression`. For improved debugging experience and better readability, this practice should be discouraged (as far as it does not affect the performance).

ariya:
Fix #1702

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -1465,7 +1465,6 @@ export class Parser {
         if (prec > 0) {
             this.nextToken();
 
-            token.prec = prec;
             this.context.isAssignmentTarget = false;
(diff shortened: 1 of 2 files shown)


ariya:
> Looks good. My only concern here is that there doesn't seem to be a way to test this not a desire to test this so it has the potential to regress in the future.

The existing regression tests (some common libraries and their expected AST and token list) are surprisingly good enough to catch any potential regressions. In fact, I messed up my earlier tweak of `parseBinaryExpression` and those tests caught the mistake.

PR #1709, opened 2017-01-03
Add type annotations to all tokenizaton-related code

### LINKED ISSUE #1705: Type annotations on all tokenization-related code
...because at the last round of refactoring and clean-up (#1440), the code related to the tokenization process (mostly in the `Scanner` class) has not received enough TLC. For consistency and readability, as well as to reduce the amount of `any` types in the code, we shall add some type annotations appropriately.

ariya:
Fix #1705


src/jsx-parser.ts
```diff
@@ -256,9 +266,12 @@ export class JSXParser extends Parser {
         }
 
         this.scanner.throwUnexpectedToken();
+
+        /* istanbul ignore next */
+        throw new Error('Unreachable');
```
mikesherov: why is this line needed?

ariya: Without the line, the TypeScript compiler will complain: _error TS2366: Function lacks ending return statement and return type does not include 'undefined'_.

@DanielRosenwasser Could this be another problem with the flow analysis? `throwUnexpectedToken()` does nothing but to throw an exception, thus the compiler should not treat this function as terminated with an undefined return value.

DanielRosenwasser: You could instead write `return `this.scanner.throwUnexpectedToken()`, which I assume returns `never`.

ariya: @DanielRosenwasser Somehow it still does not work. I think it's because `this.scanner.throwUnexpectedToken()` does not directly throw, it calls `this.scanner.errorHandler.throwError(...)` which does throw.

Strangely, if I inline the content of `this.scanner.throwUnexpectedToken()` here, then the compiler is quite happy.

DanielRosenwasser: If you add a type annotation to `throwError` of type `never`, and then have `throwUnexpectedToken` actually return the call to `throwError`, you should get it working.

ariya: @DanielRosenwasser That does the trick! Thank you so much 👍

PR #1726, opened 2017-01-10
Simplify handling the marker for node location

### LINKED ISSUE #1717: Merge the different structures to track node location
Currently tracking the location of a node requires a mash-up of various structures: MetaNode, Marker, JSXMetaNode. This is highly confusing. For improved readability, all of them should be consolidated into just one structure.

ariya:
Marker, MetaNode, JSXMetaNode are all merged into just one structure

Fix #1717

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -31,13 +31,7 @@ interface Context {
     strict: boolean;
 }
 
-interface Marker {
-    index: number;
-    lineNumber: number;
-    lineStart: number;
-}
-
-interface MetaNode {
+export interface Marker {
     index: number;
     line: number;
     column: number;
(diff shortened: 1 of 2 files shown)


src/parser.ts
```diff
@@ -159,19 +153,19 @@ export class Parser {
 
         this.startMarker = {
             index: 0,
-            lineNumber: this.scanner.lineNumber,
-            lineStart: 0
+            line: this.scanner.lineNumber,
+            column: 0,
```
mikesherov: Stray comma?

ariya: Oops, good catch. Thank you for the notice @mikesherov!

mikesherov: Maybe there's a tslint rule for this?

PR #1737, opened 2017-01-26
Create the documentation suitable for readthedocs.io

### LINKED ISSUE #1477: Use Read the Docs for the documentation
It is expected that Esprima will continue to incorporate more useful features in the forthcoming version 3.0 (as well as any future versions). The [simplistic documentation](http://esprima.org/doc) available on the web site will not be sufficient anymore. At this point, it is just easier to leverage [Read the Docs](https://readthedocs.org/) to generate and host the documentation.

Some other additional advantages:
- [per-version documentation](https://docs.readthedocs.org/en/latest/versions.html)
- [integrated search function](https://docs.readthedocs.org/en/latest/features.html#search)
- [PDF generation](https://docs.readthedocs.io/en/latest/features.html#pdf-generation)

ariya:
Currently the contents are:

  Chapter 1. Getting Started
  Chapter 2. Syntactic Analysis (Parsing)
  Chapter 3. Lexical Analysis (Tokenization)
  Appendix A. Syntax Tree Format

Fix #1477


ariya:
Note that this doesn't yet contain new ES2017 stuff (trailing commas, async await, etc) because I also want to backport it to 3.x branch to have the versioned docs (see #1477 for some details).

docs/getting-started.md
```diff
@@ -0,0 +1,129 @@
...
+The left panel is a code editor, it can accept any JavaScript source. If the source is a valid JavaScript program, the right panel shows the [syntax tree](https://en.wikipedia.org/wiki/Abstract_syntax_tree) as the result of parsing that JavaScript program. The syntax tree can be displayed in its original form (JSON, but formatted) or in its visual form (hierarchical node view).
+
+## Supported environments
+
+Since Esprima is written in JavaScript, it can run on various JavaScript environments, including (but not limited to):
+
+* Modern web browsers (the latest version of Edge, Firefox, Chrome, Safari, etc)
+* Legacy web browsers (Internet Explorer 10 or later)
+* [Node.js](https://nodejs.org/en/) v4 or later
+* JavaScript engine in Java, such [Rhino](https://developer.mozilla.org/en-US/docs/Mozilla/Projects/Rhino) and [Nashorn](https://docs.oracle.com/javase/8/docs/technotes/guides/scripting/nashorn/)
```
TheLarkInn: s/such/such as

ariya:
> I think one very useful and interactive tool that has let me explore parsers such as esprima is astexplorer.net.

The plan is to mention relevant "expand the horizon" related tools from http://esprima.org/demo/parse.html.

docs/getting-started.md
ariya: Good catch, thanks!

TheLarkInn:
Awesome. Otherwise looks great.

mikesherov:
👍

mikesherov:
Amazing.

docs/lexical-analysis.md
```diff
@@ -0,0 +1,188 @@
...
+});
+```
+
+An example run is shown in the following screenshot (the script is called `highlight.js`):
+
+![Syntax highlighting](highlight.png)
+
+The script uses the [readline module](https://nodejs.org/api/readline.html) to read the input line-by-line, collecting each line to a local string buffer. Once there is no more input, it invokes Esprima tokenizer to break the source into a list of tokens. The script only cares about identifier tokens, hence the filtering. For each token, the starting location is used to determine where to insert the [escape code](https://en.wikipedia.org/wiki/ANSI_escape_code) to change the color to cyan and the end location is used to reset the color. This is done from the last identifier token to the first identifier token, which necessities the sorting in reverse order.
+
+For a real-world syntax highlighter that has many more features, take a look at _cardinal_ (source repository: [github.com/thlorenz](https://github.com/thlorenz/cardinal). It uses a similar approach, i.e. using Esprima tokenizer to break the source into tokens and then wrap each token with a type-specific color.
```
ariya: Ugh, the mark-up for the link is busted.

docs/syntax-tree-format.md
```diff
@@ -0,0 +1,762 @@
...
+}
+
+interface ExportDefaultDeclaration {
+    type: 'ExportDefaultDeclaration';
+    declaration: Identifier | BindingPattern | ClassDeclaration | Expression | FunctionDeclaration;
+}
+
+interface ExportNamedDeclaration {
+    type: 'ExportNamedDeclaration';
+    declaration: ClassDeclaration | Function | VariableDeclaration;
```
ariya: Hmm, pretty sure I meant `FunctionDeclaration` here.

ariya:
Now available at http://esprima.readthedocs.io! 🎉

Issue #1738, opened 2017-01-28
New documentation: error


sanex3339:
Error in chapter 2
https://esprima.readthedocs.io/en/latest/syntactic-analysis.html#distinguishing-a-script-and-a-module

Source type is `string`
`> esprima.parse('answer = 42', { sourceType: 'string' });`

ariya:
Nice catch thanks @sanex3339! Interested in creating a pull request to fix it?

PR #1739, opened 2017-01-29
Fixed wrong source type in the documentation

sanex3339:
fixes https://github.com/jquery/esprima/issues/1738


ariya:
Landed, thank you 👍

PR #1740, opened 2017-01-29
Support dynamic import call

### LINKED ISSUE #1728: Dynamic import syntax support
I couldn't find an issue tracking this already, but let me know if it is a duplicate. This may well be implementable now that it is Stage 3 (https://github.com/estree/estree/pull/137).

ariya:
Fix #1728

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -680,6 +680,8 @@ export class Parser {
                         expr = this.finalize(node, new Node.ThisExpression());
                     } else if (this.matchKeyword('class')) {
                         expr = this.parseClassExpression();
+                    } else if (this.matchImportCall()) {
+                        expr = this.parseImportCall();
                     } else {
                         expr = this.throwUnexpectedToken(this.nextToken());
                     }
@@ -1216,6 +1218,8 @@ export class Parser {
             } else {
                 this.throwUnexpectedToken(this.lookahead);
             }
+        } else if (this.matchKeyword('import')) {
+            this.throwUnexpectedToken(this.lookahead);
         } else {
             const callee = this.isolateCoverGrammar(this.parseLeftHandSideExpression);
             const args = this.match('(') ? this.parseArguments() : [];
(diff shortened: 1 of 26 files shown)


src/parser.ts
```diff
@@ -1789,10 +1813,14 @@ export class Parser {
                     statement = this.parseExportDeclaration();
                     break;
                 case 'import':
-                    if (!this.context.isModule) {
-                        this.tolerateUnexpectedToken(this.lookahead, Messages.IllegalImportDeclaration);
+                    if (this.matchImportCall()) {
```
mikesherov: Oh, interesting, import can be called from scripts? Nice to know!

ariya:
> Maybe a test showing it can't be used in a newExpression?

Good point, I can definitely at that.

src/parser.ts
ariya: That's my interpretation, not sure if it's correct or not. @domenic do you have an opinion?

domenic: Yes, definitely.

ariya:
Ah, I also still forget to tackle the case such as `await import(x)`.

doplumi:
I get an error with a StackTrace that leads to

```
    parseImportDeclaration(): Node.ImportDeclaration {
        if (this.context.inFunctionBody) {
            this.throwError(Messages.IllegalImportDeclaration);
        }
```

in [parser.ts](https://github.com/jquery/esprima/blob/master/src/parser.ts#L3470-L3472), while parsing valid es2018.

It seems like `import` statements inside of functions are rejected.

doplumi:
I get an error with a stack trace that leads to

```
parseImportDeclaration(): Node.ImportDeclaration {
    if (this.context.inFunctionBody) {
        this.throwError(Messages.IllegalImportDeclaration);
    }
```

in [parser.ts](https://github.com/jquery/esprima/blob/master/src/parser.ts#L3470-L3472), while parsing valid es2018.

It seems that `import` statements inside of functions are rejected.

PR #1742, opened 2017-01-31
Fix incorrect detection of async function

ariya:
Refs #1079

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -2893,7 +2893,7 @@ export class Parser {
             const next = this.scanner.lex();
             this.scanner.restoreState(state);
 
-            match = (state.lineNumber === next.lineNumber) && ((next.type === Token.Keyword) || (next.value === 'function'));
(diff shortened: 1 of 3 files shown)


ariya:
This is not easy to test. However, it is possible to create an invalid program that triggers the code path and throws an exception. The exception will be used for the verification

src/parser.ts
```diff
@@ -2893,7 +2893,7 @@ export class Parser {
             const next = this.scanner.lex();
             this.scanner.restoreState(state);
 
-            match = (state.lineNumber === next.lineNumber) && ((next.type === Token.Keyword) || (next.value === 'function'));
+            match = (state.lineNumber === next.lineNumber) && ((next.type === Token.Keyword) && (next.value === 'function'));
```
mikesherov: Unnecessary grouping parens.

ariya: Good catch, thanks!

PR #1750, opened 2017-02-05
Fix the node of the object property key for async function

### LINKED ISSUE #1745: Inconsistent object key property with async function
Test case:
```js
> esprima.parse('x = { async 1() {} }').body[0].expression.right.properties[0].key
```

Actual:
```js
{ type: 'Identifier', name: 1 }
```

Expected:
```js
{ type: 'Literal', value: 1, raw: '1' }
```
<hr>

The reason the key needs to be _Literal_ is because that is the case for the non-async version:
```js
> esprima.parse('x = { 1() {} }').body[0].expression.right.properties[0].key
Literal { type: 'Literal', value: 1, raw: '1' }
```

ariya:
This also corrects the location for the said property key.

Fix #1745

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -840,26 +840,12 @@ export class Parser {
         let isAsync = false;
 
         if (token.type === Token.Identifier) {
-            let id = token.value;
+            const id = token.value;
             this.nextToken();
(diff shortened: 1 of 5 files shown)


mikesherov:
:+1:

PR #1754, opened 2017-02-08
Allow async arrow function to accept yield as the parameter

### LINKED ISSUE #1747: Async arrow does not accept yield as the parameter
Test case:
```js
esprima.parse('async yield => x')
```

Actual: A thrown exception _Line 1: Unexpected token yield_.

Expected: A valid syntax tree.

<hr>

Compare it with the regular arrow function:

```js
> esprima.parse('yield => 0')
Script {
  type: 'Program',
  body: [ ExpressionStatement { type: 'ExpressionStatement', expression: [Object] } ],
  sourceType: 'script' }
```

ariya:
Fix #1747

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -1678,14 +1678,16 @@ export class Parser {
             let token = startToken;
             expr = this.parseConditionalExpression();
 
-            if (token.type === Token.Identifier && (token.lineNumber === this.lookahead.lineNumber) && token.value === 'async' && (this.lookahead.type === Token.Identifier)) {
-                const arg = this.parsePrimaryExpression();
-                this.reinterpretExpressionAsPattern(arg);
-                expr = {
-                    type: ArrowParameterPlaceHolder,
-                    params: [arg],
-                    async: true
-                };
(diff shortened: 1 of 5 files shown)


src/parser.ts
```diff
@@ -1678,14 +1678,16 @@ export class Parser {
...
 
-            if (token.type === Token.Identifier && (token.lineNumber === this.lookahead.lineNumber) && token.value === 'async' && (this.lookahead.type === Token.Identifier)) {
-                const arg = this.parsePrimaryExpression();
-                this.reinterpretExpressionAsPattern(arg);
-                expr = {
-                    type: ArrowParameterPlaceHolder,
-                    params: [arg],
-                    async: true
-                };
+            if (token.type === Token.Identifier && (token.lineNumber === this.lookahead.lineNumber) && token.value === 'async') {
```
mikesherov: Might, at this point, pay clarity wise to have a `matchAsync` function.

ariya: Unfortunately, `match` family is typically for the lookahead token, not the current token. Hopefully we can specialize such a distinctive "identifier" someday with another approach.

PR #1756, opened 2017-02-14
deep clone node when node.shorthand = true

### LINKED ISSUE #1755: Should create new Identifer Node when node.shorthand = true
## TestCase:

case 1:

```
let name = 'test';
let email = ''
let user = {name, email}
```

case 2:

```
let user = {name: 'test', email: 'abc@xxx.com'}
let {name, email} = user;
```
esprima parse  code into ast as following:

```
{
    "type": "ObjectExpression",
    "properties": [
        {
            "type": "Property",
            "key": {
                "type": "Identifier",
                "name": "name"
            },
            "computed": false,
            "value": {
                "type": "Identifier",
                "name": "name"
            },
            "kind": "init",
            "method": false,
            "shorthand": true
        },
        {
            "type": "Property",
            "key": { 
                "type": "Identifier",
                "name": "email"
            },
            "computed": false,
            "value": {
                "type": "Identifier",
                "name": "email"
            },
            "kind": "init",
            "method": false,
            "shorthand": true
        }
    ]
}
```

see the above AST tree, each property, the  Node(`key`) ===  the Node(`value`) , so when eshortten change the Identifier name into shortten chars,  bug cames

## Actual output:

```
let {a, b} = user;
```

## Expected

```
let {name:b,email:c}=a;
```

so esprima should clone the value Node from the key Node, instead of refer value Node to key Node

fishbar:
fix #1755

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -904,7 +904,7 @@ export class Parser {
                     value = this.finalize(node, new Node.AssignmentPattern(id, init));
                 } else {
                     shorthand = true;
-                    value = id;
+                    value = this.finalize(node, new Node.Identifier(token.value));
                 }
             } else {
                 this.throwUnexpectedToken(this.nextToken());
@@ -1982,7 +1982,7 @@ export class Parser {
             } else if (!this.match(':')) {
                 params.push(keyToken);
                 shorthand = true;
-                value = init;
+                value = this.finalize(node, new Node.Identifier(keyToken.value));
             } else {
                 this.expect(':');
                 value = this.parsePatternWithDefault(params, kind);


michaelficarra:
LGTM.

ariya:
Looks good @fishbar! Could also extend `api-tests.js` in the same style of commit d27d440c. Thanks!

ariya:
On a second look, this change is superfluous. It is already implemented as part of PR #1520 (to fix #1519).

Thus, unfortunately I will have to revert it.

fishbar:
it's  ok, : )

PR #1770, opened 2017-02-27
Update TypeScript-related dependencies

ariya:
The minor code change is due to the more aggressive variable shadowing checks from TSLint.


ariya:
Not sure why Escope tests (part of downstream check) started to failed. It certainly has nothing to do with the changes in this pull requests. We may need to exclude Escope from the downstream tests.

PR #1771, opened 2017-02-27
Downstream test: temporarily exclude escope

ariya:
Its tests are falling apart for no particular reason.


mikesherov:
:+1:

Mike Sherov

michaelficarra:
@ariya You should report the issues you're having with escope.

ariya:
> @ariya You should report the issues you're having with escope.

Of course, I'm still investigating where the problem lies.

Issue #1713, opened 2017-01-06
Use member access modifier in the scanner


ariya:
To eliminate unintentional changes, use `public` and `private` whenever appropriate.

Reference: http://www.typescriptlang.org/docs/handbook/classes.html

PR #1779, opened 2017-03-12
Add public access modifier to ErrorHandler methods

klaiderklai:


ariya:
Thanks @klaiderklai for the your contribution! We started to add access modifiers (see #1713) particularly in a situation where it is helpful to distinguish between a set of public and private methods.
In this PR however, this is not the case since every method of _ErrorHandler_ is (bound to be) public. Adding `public` where there is not necessary any `private` counterpart doesn't add much of a value.

klaiderklai:
@ariya Ah! I see, I was going to think about that.

ariya:
I'm going to close this since we won't pursue it.

PR #1782, opened 2017-03-15
Move if conditions into switch cases

klaiderklai:
The reason to put a `switch` here was to reduce the binary expressions (i.e., `str === 'operator'`).

And I've changed the line 630, below:

```
                            str = this.source[this.index];
```

to:

```
                        str = str[0];
```

as it might reduce looking at the `this.source` String elements.

### DIFF
--- a/src/scanner.ts
+++ b/src/scanner.ts
@@ -635,7 +635,7 @@ export class Scanner {
                         if ('<>=!+-*%&|^/'.indexOf(str) >= 0) {
                             ++this.index;
                         }
-                }
+            }
         }
 
         if (this.index === start) {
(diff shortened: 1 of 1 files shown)


mikesherov:
Hi, thanks for contributing. The test suite is failing as a result of the attempted patch, and I see the branch name is `patch-2` which means you made this edit from GitHub's online editor.

I suggest cloning the repo locally, making your modifications and running the test suite to ensure no regressions, and only then, submitting a pull request with the changes. I also suggest creating an issue first, so we can discuss the merits of these changes before you implement them.

Thanks! In the meantime, I'm going to close this pull request.

klaiderklai:
@mikesherov Thanks for the help... The repository is kinda heavy, but I'll try to clone it soon again.

PR #1522, opened 2016-08-25
Update various dependencies

ariya:

PR #1794, opened 2017-03-26
Upgrade lodash

baweaver:
Upgrading Lodash to the 4.x version family. A lot of major projects are using older versions, leading to multiple versions of lodash being installed for a single project.

Tests are currently green on local running `v6.3.0` of node.


mikesherov:
Thanks for contributing. We typically update dependencies prior to a release as part of our release workflow. With that said, looks like tests are failing with this change, and also we don't use yarn so please remove the yarn.lock file.

ariya:
I've tried this in the past with no success, e.g. https://github.com/jquery/esprima/pull/1522#issuecomment-242276428. Moving away from the current version of lodash is guaranteed to break the Windows build.

Feel free to open an issue if you are willing to investigate it. Otherwise, we're stuck (unfortunately).

baweaver:
Yeah, I figured it wouldn't be that easy somehow :/

Honestly more of trying to reconcile all the disparate versions out there where possible to reduce some of the more massive bundle sizes in the wild.

I'll investigate some more on it and close this PR as a non-starter for the moment.

ariya:
In this case, it should be less of a concern since lodash is more of development dependencies (Esprima has zero run-time dependency).

PR #1797, opened 2017-03-26
Downstream test: temporarily exclude redeyed

ariya:
(while its smoke tests are being investigated)


ariya:
Obsoleted by PR #1798 (update CircleCI run to use Node.js v6).

PR #1798, opened 2017-03-26
Speculative fix to downstream tests: run them with Node.js v6.1

ariya:


mikesherov:
Interesting.

ariya:
Looks like we can get redeyed back, at the cost of documentjs (outdated, can't run on modern Node.js).
Meanwhile, Istanbul is _almost_ working again, as only one test mysteriously failed.

PR #1820, opened 2017-04-03
Improve error message upon an invalid JSX element

### LINKED ISSUE #1815: Improve error message details when encountering an invalid JSX element
Test case:
```js
esprima.parse('<a/!', { jsx: true })'
```

Expected error message: "Unexpected token !"

Actual error message: "Unexpected token ILLEGAL"

<hr>

A similar case is parsing `<a> `, the expected error message should be "Unexpected end of input", not just the generic "Unexpected token ILLEGAL".

ariya:
Instead of trying to handle it by itself, the special lexer for JSX should
simply defer to the vanilla lexer if it can't understand the upcoming token.

Fix #1815

### DIFF
--- a/src/jsx-parser.ts
+++ b/src/jsx-parser.ts
@@ -15,8 +15,7 @@ interface MetaJSXElement {
 
 const enum JSXToken {
     Identifier = 100,
-    Text,
-    Unknown
+    Text
 }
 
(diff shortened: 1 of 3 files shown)


mikesherov:
👍

Issue #1814, opened 2017-03-31
Esprima is failing to parse `ThrowStatement` with newlines on template literals


magano:
## Summary

Esprima is failing to parse `ThrowStatement` with newlines on template literals

## Configuration

* esprima 3.1.3

## Description

Parsing a `ThrowStatement` with a template literal including a newline displays the following error: `Error: Line 1: Illegal newline after throw`.
The following sample code is syntatically valid but throws an error while parsing.
```javascript
throw `new Error: Error on loop
For more details:
http://example.com`
```
This issue was also tested on the latest released version as a npm package

## Steps to Reproduce

* Go to http://esprima.org/demo/parse.html
* Add the code above
* Check the error.

## Actual Results

`Error: Line 1: Illegal newline after throw`

## Expected Results

```
{
    "type": "Program",
    "body": [
        {
            "type": "ThrowStatement",
            "argument": {
                "type": "TemplateLiteral",
                "quasis": [
                    {
                        "type": "TemplateElement",
                        "value": {
                            "raw": "new Error: Error on loop\nFor more details:\nhttp://example.com",
                            "cooked": "new Error: Error on loop\nFor more details:\nhttp://example.com"
                        },
                        "tail": true
                    }
                ],
                "expressions": []
            }
        }
    ],
    "sourceType": "script"
}
```

mikesherov:
Thanks for contributing! We'll take a look soon!

PR #1821, opened 2017-04-03
Fix ThrowStatement parsing for multiline TemplateLiterals

magano:
Currently esprima fails to parse multiline `TemplateLiterals`

Issue reported: https://github.com/jquery/esprima/issues/1814

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -2583,12 +2583,26 @@ export class Parser {
     parseThrowStatement(): Node.ThrowStatement {
         const node = this.createNode();
         this.expectKeyword('throw');
+        const hasLineTerminator = this.hasLineTerminator;
 
-        if (this.hasLineTerminator) {
-            this.throwError(Messages.NewlineAfterThrow);
+        let argument;
+        try {
+            argument = this.parseExpression();
+        } catch (e) {
+            if (!hasLineTerminator) {
+                throw e;
+            }
         }
 
-        const argument = this.parseExpression();
+        if (hasLineTerminator) {
+            const shouldThrow = !argument ||
+                (argument && argument.type !== 'TemplateLiteral');
+
+            if (shouldThrow) {
+                this.throwError(Messages.NewlineAfterThrow);
+            }
+        }
+
         this.consumeSemicolon();
 
         return this.finalize(node, new Node.ThrowStatement(argument));
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -2584,6 +2584,7 @@ export class Parser {
         const node = this.createNode();
         this.expectKeyword('throw');
         const hasLineTerminator = this.hasLineTerminator;
+        const source = this.scanner.source;
 
         let argument;
         try {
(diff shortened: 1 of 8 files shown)


magano:
Forgot to mention that this changes the error line for the newline errors. I kept it that way and changed the tests accordingly.
That's mainly because I couldn't find a good way of pointing to the previous context to get the line.
Anyway please let me know if you need further details or any changes to be applied.

src/parser.ts
```diff
@@ -2583,12 +2583,26 @@ export class Parser {
...
-        if (this.hasLineTerminator) {
-            this.throwError(Messages.NewlineAfterThrow);
+        let argument;
+        try {
+            argument = this.parseExpression();
+        } catch (e) {
+            if (!hasLineTerminator) {
+                throw e;
+            }
+        }
```
ariya: Hmm, what's actually the logic behind this try-catch?

test/fixtures/statement/throw/migrated_0003.js
```diff
@@ -0,0 +1,2 @@
+throw `Error:
+Generic Error`;
```
ariya: Since this is about template literal, the more proper place for this test is in `test/fixtures/ES6/template-literals/`. Also, use a more descriptive name.

src/parser.ts
```diff
@@ -2583,12 +2583,26 @@ export class Parser {
...
+
+        if (hasLineTerminator) {
+            const shouldThrow = !argument ||
+                (argument && argument.type !== 'TemplateLiteral');
+
+            if (shouldThrow) {
+                this.throwError(Messages.NewlineAfterThrow);
+            }
         }
```
ariya: I don't think this logic handles the following case (expected to still thrown an exception):
```
throw
`foo
bar`
```

src/parser.ts
magano: The idea was to keep the previous behaviour were the line terminator errors were caught before any error coming from the parseExpression.

test/fixtures/statement/throw/migrated_0003.js
magano: Will move it to the template-literals.
You mean the filename?

src/parser.ts
ariya: But in the current implementation, the exception was thrown inside this function itself.
Could you illustrate an example that triggers the behavior that you want to keep?

src/parser.ts
```diff
@@ -2583,12 +2583,34 @@ export class Parser {
...
+                throw e;
+            }
+        }
+
+        if (hasLineTerminator) {
+            const isTemplateLiteral =
+                argument && argument.type === 'TemplateLiteral';
+            const hasLineTerminatorBeforeTemplate =
+                isTemplateLiteral &&
+                source.trim().indexOf('\n') < source.trim().indexOf('`');
```
ariya: This is almost guaranteed to be wrong, `source` holds the _entire_ input to parsed. You can't just run `trim` and look for a particular character. What if there are multiple template literals in the input?

src/parser.ts
```diff
@@ -2583,12 +2583,34 @@ export class Parser {
...
+
+            if (
+                !argument ||
+                !isTemplateLiteral ||
+                hasLineTerminatorBeforeTemplate
+            ) {
```
ariya: Nitpick: look at various other condition branches in the code base, we typically do not use this code style (per-line conditional check).

ariya:
This pull request still needs some more tweaks. Feel free to reopen it once there is a new change.

PR #1832, opened 2017-05-10
Fix the "tokens" and "comment" argument names

syllant:
Were specified as "token" and "comments" in the description of the parse method


ariya:
Thank you @syllant 👍

PR #1835, opened 2017-05-20
Added some useful links to docs.

gs-akhan:
Hi, 
I found these links really helpful in learning concepts of ASTs. Adding them to docs too.

Thanks

### DIFF
--- a/doc/videos.html
+++ b/doc/videos.html
@@ -82,6 +82,22 @@
   <!-- Main content -->
   <div class="row">
     <div class="eight columns">
+      <h5>Harnessing The Power of Abstract Syntax Trees.</h5>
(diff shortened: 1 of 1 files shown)


ariya:
Thank you for the PR @gs-akhan!
These videos are all very good. However, they mostly cover the use of Babel (e.g. creating a plugin, etc), which has some important differences (e.g. AST format) than Esprima (and its related tools). I'm afraid that adding them here will create unnecessary confusion for the reader.
I recommend that you propose adding those videos to Babel website instead! /cc @hzoo

hzoo:
Cool. Haven't found a good place/opportunity for it but https://github.com/babel/babel.github.io/issues/27 - can be easy for such as list to get too long or out of date.

PR #1843, opened 2017-07-25
Fix incorrect YieldExpression  in object methods

### LINKED ISSUE #1834: yield should be parsed as identifier in object methods
### Steps to reproduce

```js
function *a(){({b(){yield}})}
function *a(){({get b(){yield}})}
function *a(){({set b(c){yield}})}
```

### Expected output
In each of these cases `yield` is parsed as `YieldExpression`.

### Actual output
It should be parsed as regular `Identifier` because object method is a separate function on its own.

Kingwl:
Fix #1834

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -760,7 +760,7 @@ export class Parser {
         const node = this.createNode();
 
         const previousAllowYield = this.context.allowYield;
-        this.context.allowYield = false;
+        this.context.allowYield = true;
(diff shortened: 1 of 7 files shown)


Kingwl:
seems ci is broken

ariya:
Looks like AssetGraph requires Node.js > 6 and hence why the downstream CI failed.

PR #1845, opened 2017-07-28
Downstream test: disable AssetGraph since it needs Node.js > 6.

ariya:
See https://github.com/assetgraph/assetgraph/commit/07242b10. We'll figure out how to adapt to this.


mikesherov:
:+1:

Mike Sherov

PR #1846, opened 2017-07-31
Fix Negative Column in Binary Expression String Concatenation

### LINKED ISSUE #1834: yield should be parsed as identifier in object methods

### LINKED ISSUE #1844: Negative Column Number for Start of Binary Expression in String Concatenation with Escaped Newlines
Sorry for the long title; I wasn't sure how to summarize it better!

Given a string expression with escaped newlines and a string concatenation, Esprima reports a negative column number for the string concatenation.

Example:

```javascript
'\
' + bar
```

### Steps to reproduce

```js
var tree = esprima.parse("'\\\n' + bar", {loc: true});
// Print the LOC for the binary expression
console.log(tree.body[0].expression.loc);
```

### Expected output

```
{ start: { line: 1, column: 0 }, end: { line: 2, column: 7 } }
```

### Actual output

```
{ start: { line: 2, column: -3 }, end: { line: 2, column: 7 } }
```

### Relevant references

N/A

### Hypothesis

It appears that the start location would be correct if you went back 3 characters from the beginning of line 2. Perhaps Esprima assumed that everything would have been on the same line?

### Impact

The source maps generated using this information are rejected by source map libraries (like Mozilla's [source-map](https://github.com/mozilla/source-map)) because negative column numbers are nonsensical. The source-to-source transformation I've written needs to take this bug into account, and rewrite any nodes in the AST with invalid locations.

Kingwl:
Fix #1844 

~~OMG...i ref a incorrect issue~~


Kingwl:
ping @ariya   
ci seems still broken: `operation not permitted`

ariya:
Looks like the CI problem (that got solved) was due to a quick hitch in AppVeyor...

src/parser.ts
```diff
@@ -1524,12 +1530,16 @@ export class Parser {
             // Final reduce to clean-up the stack.
             let i = stack.length - 1;
             expr = stack[i];
-            markers.pop();
+
+            let lastMarker = markers.pop();
             while (i > 1) {
-                const node = this.startNode(markers.pop());
+                const marker = markers.pop();
+                const lastLineStart = lastMarker && lastMarker.lineStart || 0;
```
ariya: The `|| 0` should not be necessary. Is there a particular reason for that?

test/fixtures/expression/binary/migrated_0018.js
```diff
@@ -0,0 +1,2 @@
+'\
```
ariya: Let's have the file named something else than `migrated_0018` (such a pattern was due a migration from the old test suite).
I suggest `multiline_string` or something similar.

ariya:
SGTM @Kingwl, except for the minor issues. Thank you!

ariya:
SGTM @Kingwl, except for the minor issues. Thank you!

Kingwl:
@ariya fix those later...my cat need me

src/parser.ts
Kingwl: done

test/fixtures/expression/binary/migrated_0018.js
Kingwl: done

ariya:
LGTM @Kingwl. Thank you!

PR #1852, opened 2017-09-12
Migrate to CircleCI 2.0

serima:
## Summary

- CircleCI runs speed up.
  - before 4:50 https://circleci.com/gh/jquery/esprima/677
  - after 2:58 https://circleci.com/gh/serima/esprima/tree/feature%2Fcircleci-20
    - If `node_modules` is cached it will be shortened by another 30 seconds
- Run parallel tests with node 6 and 8.


ariya:
Hi @serima, thanks for looking into this. Can you do me a favor? Please just combine the two commits into a single one. Also, this Circle CI is being used to run [downstream tests](http://esprima.org/test/ci.html) which is known to be sensitive to the Node.js version, hence for maximum safety we shouldn't run it with Node.js v8.

serima:
@ariya Okay. Thanks.

serima:
@ariya  Since we did not need to use workflow, changed it to sequential processing and rebased.
Please review it again :)

ariya:
LGTM, thank you @serima 👍

PR #1892, opened 2017-11-30
Run every benchmark cycle in a fresh web worker

### LINKED ISSUE #1860: [Speed Comparison] Avoiding JIT warmup
So, I was adding the new [Cherow parser](https://github.com/cherow/cherow) to the comparison benchmark, and I noticed that the results didn't match up with my unscientific benchmark "in the wild", which turned out much slower. I believe this is because the benchmark calls the parser many many times, giving the JIT compiler a lot of opportunity to optimize the code, while I only parse a few scripts, and all of them only once.

So, I modified the benchmark to create a new Web Worker for every benchmark cycle, in an attempt to make the code run in a fresh instance of the JIT compiler. (While still only measuring the parse time.) And sure enough, the parse times went up by around 2x, and closer together.

Do you think this is a fair(er) benchmark (or do you think warm JIT is more realistic for most use cases)? (We could also have both, of course.) It might also be useful for the (non-comparison) [benchmark suite](http://esprima.org/test/benchmarks.html).

You can see the (still rough) code [here](https://github.com/twiss/esprima/commit/60ab89101b3d70f974ee64e82e7808cdf6d1de5e), and run the benchmark [here](http://rawgit.com/twiss/esprima/gh-pages/test/compare.html).

([More discussion on Cherow's issue tracker](https://github.com/cherow/cherow/issues/15))

twiss:
Fix #1860.

### DIFF
--- a/test/compare.html
+++ b/test/compare.html
@@ -8,7 +8,6 @@
   <meta name="viewport" content="width=device-width" />
   <link rel="stylesheet" href="../assets/style.css">
   <link rel="stylesheet" href="../assets/foundation/foundation.min.css">
-  <script src="../esprima.js"></script>
   <script src="compare.js"></script>
   <script src="3rdparty/benchmark.js"></script>
   <script src="3rdparty/XMLHttpRequest.js"></script>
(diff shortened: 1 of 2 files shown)


ariya:
LGTM. Thank you @twiss 👍

PR #1915, opened 2018-02-21
Update identifier data to Unicode v10

mathiasbynens:


ariya:
Looks good indeed, will merge soon. Thanks @mathiasbynens :+1

PR #1924, opened 2018-04-01
Recognize Token.StringLiteral as a ReturnStatement argument.

mfitz3:
The return statement currently fails to parse the return argument for multiline strings.

Multiline strings are handled correctly for variables: http://esprima.org/demo/parse.html?code=var%20a%20%3D%20'hello%20%5C%0A%09world'%3B

But they are not properly handled as return statement arguments: http://esprima.org/demo/parse.html?code=function%20a()%20%7B%0A%20%20return%20'hello%20%5C%0A%20%20%09world'%3B%0A%7D

The ReturnStatement argument is null, but it should be: "hello   \tworld"

This is especially troublesome when applied to libraries like Angular Material which rely on multi-line strings (https://github.com/angular/material/blob/5320d225ee5ec72153aab9b670cd358c07313d5a/src/components/autocomplete/js/autocompleteDirective.js#L303).

This fix is similar to the one from #1882 for Token.Template.

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -2453,6 +2453,7 @@ export class Parser {
 
         const hasArgument = (!this.match(';') && !this.match('}') &&
             !this.hasLineTerminator && this.lookahead.type !== Token.EOF) ||
+            this.lookahead.type === Token.StringLiteral ||
             this.lookahead.type === Token.Template;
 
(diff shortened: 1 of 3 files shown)


mfitz3:
Re-sending as #1925 in order to sign CLA.

PR #1930, opened 2018-05-18
Downstream test: patch Recast to ignore its TypeScript tests.

ariya:
These tests often fail spuriously, sometimes in conjuction with GraphQL,
so let's isolate us from them.


ariya:
Closing this in favor of #1931.

Issue #1699, opened 2016-12-26
Possibly wrong tests detected


icefapper:
Hello
while stress-testing another parser with esprima, I came across a number of tests that pass while they should fail, or vice versa; some of them pass as they are expected to, but they are possibly marking the locations wrongly:

```sh
test/test-esprima/fixtures/ES6/binding-pattern/array-pattern/for-let-let
test/test-esprima/fixtures/ES6/binding-pattern/object-pattern/for-let-let
```
They tests above are passing, even though their binding patterns require initilizers.

```sh
test/test-esprima/fixtures/ES6/export-declaration/export-named-keyword-as-specifier
test/test-esprima/fixtures/ES6/export-declaration/export-named-keyword-specifier
```
The tests above are passing, even though an exported name must be an `IdentifierReference` when `[Local]` is not actually present in the production.

```sh
test/test-esprima/fixtures/ES6/yield/invalid-yield-strict-function-declaration 
test/test-esprima/fixtures/ES6/yield/invalid-yield-strict-function-expression 
```
In a similar issue filed a while back, these turned out to be valid syntax -- 'use strict' affects the `params` and the `body` of a function, not `BindingIdentifier`s they might have.

```sh
test/test-esprima/fixtures/ES6/yield/yield-generator-arrow-default
test/test-esprima/fixtures/ES6/yield/yield-generator-default-parameter
```
yield is not allowed inside an arrow's (or a function's) parameters

```sh
test/test-esprima/fixtures/ES6/template-literals/literal-escape-sequences
```
`\u0` is actually an invalid escape sequence.

```sh
test/test-esprima/fixtures/invalid-syntax/migrated_0201 
test/test-esprima/fixtures/invalid-syntax/migrated_0202
test/test-esprima/fixtures/invalid-syntax/migrated_0205 
test/test-esprima/fixtures/invalid-syntax/migrated_0206
test/test-esprima/fixtures/invalid-syntax/migrated_0208
test/test-esprima/fixtures/invalid-syntax/migrated_0234
test/test-esprima/fixtures/invalid-syntax/migrated_0235
test/test-esprima/fixtures/invalid-syntax/migrated_0236
```
they are actually valid -- just like this test: `fixtures/ES6/yield/invalid-yield-strict-function-declaration`.

```sh
test/test-esprima/fixtures/es2017/async/methods/invalid-async-class-getter ## valid
test/test-esprima/fixtures/es2017/async/methods/invalid-async-class-setter ## valid
[...]

PR #1965, opened 2019-03-15
Fixed miscalculated 'computed' field

MikaelMayer:
This PR fixes the last bug of #1699  "`computed` must be true, but it is not."
I added the missing line in `src/parser.ts`

I know that this is the correct fix because I also fixed it in my [forked version of Esprima](https://github.com/MikaelMayer/esprima) where I support all concrete syntax elements in nodes, as well as a basic `.unparse()` method on every node.
All tests pass.

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -3240,6 +3240,7 @@ export class Parser {
                 if (punctuator !== ':' && punctuator !== '(' && punctuator !== '*') {
                     isAsync = true;
                     token = this.lookahead;
+                    computed = this.match('[');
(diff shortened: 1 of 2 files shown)


donnagf:
@MikaelMayer Don't expect this PR to be merged. This repo is stall and so is the author it seems. 7 months and the author doesn't even look at your PR :) Well. There are more than 1 year old PR's he didn't look into at all either.

ariya:
Thank you @MikaelMayer :+1:

PR #1969, opened 2019-04-29
Update README fix typo

francolaiuppa:
The `esprima.parseScript` call is outdated, it's now called `esprima.parse`. I've updated README.md to reflect this.

Hope it helps! This is my first contribution, so feel free to modify it if it does not meet the contribution guidelines.


ariya:
Hi @francolaiuppa, thanks for the PR! 

The example is already correct: `parse` is old and it should not be used. For more info, refer to https://esprima.readthedocs.io/en/4.0/syntactic-analysis.html#distinguishing-a-script-and-a-module.

PR #1978, opened 2019-07-02
Update identifier parsing per Unicode v12.1.0

mathiasbynens:
ECMAScript follows the latest version of the Unicode Standard. Unicode v12.1.0 was officially released on May 7th, 2019. This patch updates Esprima to support the new identifier characters.

Ref. https://github.com/tc39/ecma262/issues/1530.


ariya:
Merged. Thank you @mathiasbynens!

PR #1994, opened 2019-09-11
Fix Unexpected ILLEGAL token with function literal in template strings

### LINKED ISSUE #1991: Unexpected token ILLEGAL with function literal in template literal
### Steps to reproduce

Attempting to parse this code causes an error:

```js
`${function() { let {x} = y; }}`
```

This is a reduced test case from a real app. We were using a string like this to create the source code for a web worker.

### Expected output

This should be able to parse fine. The code parses fine with other parsers and works fine in the Chrome Developer Tools console.

### Actual output

I get this instead:

```
Error: Line 1: Unexpected token ILLEGAL
```

JosephPecoraro:
Fixes #1991

### DIFF
--- a/src/scanner.ts
+++ b/src/scanner.ts
@@ -50,6 +50,7 @@ interface ScannerState {
     index: number;
     lineNumber: number;
     lineStart: number;
+    curlyStack: string[];
 }
 
 export class Scanner {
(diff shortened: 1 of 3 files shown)


JosephPecoraro:
@ariya Small Esprima bug fix! Are you available to review?

PR #1999, opened 2019-10-01
Support preserving single-line comment style

GuyLewin:
Support preserving the single-line comment used in parsed Javascript code, so that escodegen can use the right style ('//', '', '/*') to rebuild the tree.
I would have preferred to replace "type" in comment to be this prefix, but that would obviously break the API.

This solves issue #2003


GuyLewin:
I'm aware of the unsuccessful tests and fixing the UTs.

GuyLewin:
It looks like Firefox crashed in AppVeyor build according to the logs

GuyLewin:
@ariya whenever you have time, this PR is to keep the comment style while parsing. Thanks!

donnagf:
@GuyLewin This repo is stall. Don't expect too much. PR's that is over 1y old and he never looked into them or merged. The author was active last 22 days ago (see his profile) and promised to follow up on other PR's. Didn't happen.

I just don't say you shouldn't expect too much. Even the ES2018 feature implementation that was scheduled to mid last year is delayed. See open project for reference.

And the author doesn't like that I use the word "stall" so be aware. But just look around on the repo and you understand.

ariya:
Hi @GuyLewin, thanks for the PR! However, could you open an issue first? Since it changes the API, we need to be a bit more formal in discussing the API, any potential performance impact, doc change, etc. Thank you!

GuyLewin:
@ariya thanks! Created an issue ( #2003 ) and updated the API documentation in this PR.
Please have a look at the implementation dilemma I wrote in the issue, whether to split `LineComment` into 3 types (which will break API) or to add `style` property to all comments (that's how I chose to implement).

Thanks again

ariya:
After some more deliberation, I think it's better to do this as either a post-processing step or by leveraging the [syntax node delegate]https://esprima.readthedocs.io/en/latest/syntactic-analysis.html#syntax-delegate).
We will likely keep comment attachment as a backward compatibility feature and would not change nor extend it for any foreseeable future.
Thank you very much!

Issue #2005, opened 2019-11-03
For{In,Of}Statement are missing VariableDeclaration as possible value for `left` in documentation


wilsonzlin:
The `ForInStatement` and `ForOfStatement` interfaces can also have `VariableDeclaration` as the value for the `left` property, as specified by the spec at https://www.ecma-international.org/ecma-262/10.0/index.html#sec-for-in-and-for-of-statements and demonstrated by running the following code:

```javascript
esprima.parseScript('for (const a in b) {}');
esprima.parseScript('for (const a of b) {}');
```

The code itself works fine, it's simply missing in the documentation, which says it can only have a value of type `Expression`.

PR #2007, opened 2019-11-08
#2005 [Documentation] Add missing VariableDeclaration to For{In,Of}Statement types

wilsonzlin:
Issue: #2005

The documentation only specifies Expression as the value for the left property of ForInStatement and ForOfStatement. This adds the VariableDeclaration type to the left property in the documentation.


ariya:
Merged. Thank you @wilsonzlin 👍

PR #2009, opened 2020-01-01
Migrate from tslint to eslint and tsfmt to prettier.

calvinf:
Because `tslint` is deprecated, this commit migrates from `tslint` to `eslint` with `typescript` and `prettier`.  There is a new `lint` script that runs it, and prettier is set up to run on the generated code so it's in the format expected rather than relying on tsfmt.

As package-lock.json has not previously been generated here, it's now generated for the first time.

`eslint` and `prettier` have been run against relevant files.


calvinf:
It's not clear to me why the `test-262` tests are failing on circleci and if that's affected by the particular node version that's being run or if something else that changed in this PR is causing the issue.

ariya:
Hi @calvinf, thanks for the help!
IMHO there's too much going on in one PR like this, it's hard to understand and fix the breakage.

calvinf:
I'll make some smaller PRs.

PR #2011, opened 2020-01-01
Update CI node versions and increase minimum Node version to 8.

calvinf:
Before working on updating dependencies, it makes sense to ensure esprima is tested against currently supported node versions.  Node 8 is nearly deprecated so making that the minimum version; bumping minimum CI version to Node 10.

Add package lock to support using `npm ci` for a CI-optimized module install path.  This broke one test on appveyor for unknown reasons so continuing to use `npm install` there.

Update test-262-whitelist against Node 10.


ariya:
Thank you @calvinf! I reckon the lock file could have been a separate PR, but this'll do for now.

Issue #2010, opened 2020-01-01
Migrate from tslint to eslint


calvinf:
Because `tslint` is deprecated, I propose to migrate to `eslint` with typescript support and use prettier to handle the code style rules.

### Steps to reproduce
If you view the [tslint repo](https://github.com/palantir/tslint) they have a message stating "TSLint will be deprecated some time in 2019".  As such, updating this repo to use eslint instead.

### Proposed change

* Upgrade to latest TypeScript version.
* Add relevant eslint and prettier dependencies.
* Migrate the rules to be close to the previously applied tslint code logic rules.
* Turn off code style rules and rely on prettier for code style.

### Relevant references

My go to guide for setting up typescript, eslint, prettier: https://www.robertcooper.me/using-eslint-and-prettier-in-a-typescript-project

### Pull request

* https://github.com/jquery/esprima/pull/2009

calvinf:
#2009 is canceled; I'm going to re-work this into some smaller PRs.

PR #2013, opened 2020-01-04
Migrate from TSLint to ESLint

zweimach:
I made this changes last year but haven't had the courage to open a PR.
Following issue #2010 made by @calvinf, I encourage myself to open this.
Because TSLint was deprecated last year, I made this commit to migrates from TSLint to ESLint with support for TypeScript.
I already fixed some of the lint rules violation created by running ESLint against existing files.
Please kindly review my PR. Thank you.


src/esprima.ts
```diff
@@ -94,11 +94,10 @@ export function parseScript(code: string, options, delegate) {
 export function tokenize(code: string, options, delegate) {
     const tokenizer = new Tokenizer(code, options);
 
-    let tokens;
-    tokens = [];
+    const tokens: any = [];
 
     try {
-        while (true) {
+        for (; ;) {
```
ariya: Can we put a rule option that permits `while (true)`?
IMO `for(;;)` is far less readable (it does not telegraph the semantic of "forever").

ariya:
Thanks @zweimach! Can you rebase it on the latest master.
(Note: at some point,  better to squash this PR before it is merged)

zweimach:
@ariya feel free to squash those commits, and thanks for the review.
Glad to be of help.

src/scanner.ts
```diff
@@ -1109,7 +1110,7 @@ export class Scanner {
                 .replace(/\\u\{([0-9a-fA-F]+)\}|\\u([a-fA-F0-9]{4})/g, ($0, $1, $2) => {
                     const codePoint = parseInt($1 || $2, 16);
                     if (codePoint > 0x10FFFF) {
-                        self.throwUnexpectedToken(Messages.InvalidRegExp);
+                        this.throwUnexpectedToken(Messages.InvalidRegExp);
```
ariya: Hmm, why is this changed?

src/jsx-parser.ts
```diff
@@ -35,20 +35,23 @@ function getQualifiedElementName(elementName: JSXNode.JSXElementName): string {
     let qualifiedName;
 
     switch (elementName.type) {
-        case JSXSyntax.JSXIdentifier:
+        case JSXSyntax.JSXIdentifier: {
```
ariya: Is there also a rule/option somewhere that minimize this unnecessary change?

src/scanner.ts
zweimach: That is a recommended lint rule of [@typescript-eslint/eslint-plugin](https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/eslint-plugin),
You can read about that on [@typescript-eslint/no-this-alias](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-type-alias.md),
and here we use an arrow function expression so aliasing `this` is pretty much useless.

src/jsx-parser.ts
zweimach: No [there isn't](https://eslint.org/docs/rules/no-case-declarations). Do you want me to revert that one?

src/scanner.ts
ariya: That makes sense! Thank you for the clarification.

src/jsx-parser.ts
ariya: Yes, please!

package.json
```diff
@@ -75,11 +77,11 @@
   ],
   "scripts": {
     "check-version": "node test/check-version.js",
-    "tslint": "tslint src/*.ts",
+    "lint-code": "eslint src/*.ts",
```
calvinf: I'd call it `lint`, but that's personal preference.

zweimach: I choose `lint-code` because there is a script called `format-code`.
@ariya what do you think?

ariya:
Merged. Thank you @zweimach!

Issue #2000, opened 2019-10-09
super() in non derived class


peeyush18:
super() in non derived class constructors should throw a syntax error

### Steps to reproduce

```js
esprima.parse('class A{
 constructor(){
  super(); 
 }
}')
```
### Expected output

throw an syntax error.
in firefox
```js
 SyntaxError: super() is only valid in derived class constructors
```
in chrome
```js
Uncaught SyntaxError: 'super' keyword unexpected here
```

PR #2047, opened 2020-11-28
Fix super() in non derived class

eryue0220:
This pr solved the #2000 problem.


test/test-262-whitelist.txt
```diff
@@ -17518,4 +17514,12 @@ test/language/statements/class/async-gen-method-static/forbidden-ext/b2/cls-decl
 test/language/statements/class/async-gen-method-static/forbidden-ext/b2/cls-decl-async-gen-meth-static-forbidden-ext-indirect-access-prop-caller.js(default)
 test/language/statements/class/async-gen-method-static/forbidden-ext/b2/cls-decl-async-gen-meth-static-forbidden-ext-indirect-access-own-prop-caller-value.js(default)
 test/language/statements/class/elements/syntax/valid/grammar-special-prototype-async-gen-meth-valid.js(default)
-test/language/statements/class/elements/syntax/valid/grammar-special-prototype-async-gen-meth-valid.js(strict mode)
\ No newline at end of file
+test/language/statements/class/elements/syntax/valid/grammar-special-prototype-async-gen-meth-valid.js(strict mode)
```
ljqx, line 17517: these tests are broken.

In general, `super()` is not allowed in non derived class, but `super.xxx` is allowed.

ljqx (changes requested):

test/test-262-whitelist.txt
eryue0220: OK, I will check that.

eryue0220: I run the test in the local env, it can pass.

ljqx: Then you can remove these newly added tests from this whitelist file.

eryue0220: done.

jogibear9988:
@eryue0220 

I think this test is still broken:

test/language/statements/class/syntax/class-body-method-definition-super-property.js
see: https://github.com/tc39/test262/blob/main/test/language/statements/class/syntax/class-body-method-definition-super-property.js

PR #2048, opened 2020-12-05
Support for ES2020 Optional chaining

### LINKED ISSUE #2023: ES2020 Optional Chaining
### Syntax

```js
if (foo?.bar?.baz) {
    console.log('baz exists on foo.bar')
}
```

Which is more-or-less equivalent to the following pre-ES2020 snippet.

```js
if (foo && foo.bar && foo.bar.baz) {
    console.log('baz exists on foo.bar')
}
```

MDN does a fantastic job explaining more thoroughly: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

### Spec

TC39:
https://tc39.es/ecma262/#prod-OptionalExpression
https://github.com/tc39/proposal-optional-chaining

### Tests

https://github.com/tc39/test262/tree/master/test/language/expressions/optional-chaining

ljqx:
This PR fixes #2023.

Several changes are made:

* The new nodes (properties) defined in [ESTree spec](https://github.com/estree/estree/blob/master/es2020.md#chainexpression), including `ChainExpression`, `ChainElement` (`optional` property for `MemberExpression` and `CallExpression`)
* Scanner logic for `?.`, which cannot follow a digit.
* Parser logic for `?.`
  * in CallExpression, computed MemberExpression, or static MemberExpression
  * Optional chaining cannot be assignment target
  * Optional chaining cannot work with template string (Note the error message is different for direct chaining `a?.'b'` with template string or indirect `a?.b'c'`, here apostrophe is used as backtick)
  * `ChainExpression` is only inserted when in the chain there is optional chaining
* Tests are updated
  * Unit tests fixtures and integration tests fixtures need to be updated to add the new `optional: false` property for `MemberExpression` and `CallExpression`
  * Test 262 whitelist is updated (Note `test/language/expressions/optional-chaining/call-expression-super-no-base.js` is whitelisted because of #2000 
  * api tests is updated to include the new node type


JingleH:
When can this be merged? It would be really helpful! Thanks!

lahma:
@ljqx has done great work with the new features,  kudos! Let's give @ariya some time to check it,  I know that there can be a lot going on, like work and family.

6uliver:
When could you review and merge this PR since it's a blocking issue for other packages which are using esprima like this one: https://github.com/i18next/i18next-scanner/issues/205

Can I help you somehow to speed up the process?

jogibear9988:
@ljqx @ariya any news to this, when this could be merged?

johnjbarton:
We get a lot of complaints now about `esprima` failing on optional chaining.

ljqx:
Will be reviewed in this weekend, thanks @ariya !

ariya (approved):
Excellent work :1st_place_medal:

PR #2049, opened 2020-12-05
Support for ES2018 template literal revision

### LINKED ISSUE #1587: ES2017: Revised Template Literal
(to be written)

References:
* The proposal: https://tc39.github.io/proposal-template-literal-revision/
* Change to the spec: ?
* ESTree for ES2017: https://github.com/estree/estree/blob/master/es2017.md

ljqx:
This PR fixes #1587.

Several changes are made:
* TemplateElementValue type is updated, `cooked` now can be `null`. See [ESTree spec](https://github.com/estree/estree/blob/master/es2018.md#template-literals).
* `parseTemplate*` family is updated to accept options `isTagged`, matched with the [spec update](https://tc39.es/proposal-template-literal-revision/#sec-template-literals). This is mainly for emitting [early errors](https://tc39.es/proposal-template-literal-revision/#sec-static-semantics-template-early-errors) only for non-tagged cases.
* Forbidden `\8` and `\9` which should not be allowed in non-tagged cases, while before this PR they are always allowed.
* The scanner now doesn't emit errors for [`NotEscapeSequence`](https://tc39.es/proposal-template-literal-revision/#prod-NotEscapeSequence), parser would do the work. Matched with [the spec](https://tc39.es/ecma262/#sec-static-semantics-template-early-errors), where early errors are described in `Expressions` section, instead of `Lexical Grammar` section. As we need to know if it's tagged or not, it's better to be in parser instead of the scanner.
* Cases like `\u0` is previously considered valid, but actually it's not. Here we fixed the issue too.
* Unit tests and test 262 whitelist are updated too.


ariya:
Merged. Thank you @ljqx!

PR #2052, opened 2020-12-06
Support for ES2020 import.meta

### LINKED ISSUE #2051: ES2020 import.meta
### Steps to reproduce

```js
esprima.parse('import.meta')
```

### Expected output

A syntax tree.

### Actual output

>> Error: Line 1: Unexpected token .

### Relevant references

Proposal: https://github.com/tc39/proposal-import-meta
Spec: https://tc39.es/proposal-import-meta/
ESTree: https://github.com/estree/estree/blob/master/es2020.md#metaproperty

ljqx:
This PR fixes #2051 

* when we met `import.meta` in `parseStatementListItem`, we treat it as part of a `Statement`
* when we met `import.meta` in `parsePrimaryExpression`, we parse it. 
  * if not in module, we emit an error as `import.meta` can only be used in module
  * if in module, we construct a `MetaProperty`, which cannot be assignment target.

Though `meta` is not a keyword, here it cannot be unicode escape. See [test 262](https://github.com/tc39/test262/blob/main/test/language/expressions/import.meta/syntax/escape-sequence-meta.js).


ariya:
Thank you @ljqx :+1

lahma:
@ljqx I see three tests failing in main branch, something wrong with tests or another problem maybe?

lahma:
@ariya @ljqx I've created https://github.com/jquery/esprima/pull/2099 to fix build, other PRs might require rebase too.

PR #2058, opened 2020-12-07
infrastructure: update "engines" to node>=10.9

### LINKED ISSUE #2056: infrastructure: update "engines" to node>=10.4
Last update of "engines" is about 1 year ago: https://github.com/jquery/esprima/commit/3d182f16403b6efad66acfd92eb46022728bee56

Now Node 10 is the oldest [LTS in maintenance](https://nodejs.org/en/about/releases/).

For adding support of BigInt #1988 and #2035, it's better that the environment has support of BigInt. Or the `value` of `BigIntLiteral` [would be `null`](https://github.com/estree/estree/blob/master/es2020.md#bigintliteral) for Node < 10.4, which increases maintenance cost.

ljqx:
This PR fixes #2056 

* package.json `engines` is updated to node>=10.9
* github actions are updated to by default use 14.x instead of 12.x (or 10.x)
* move node 8.x test runs to legacy, and remove 6.x from legacy
* <del>ts compile target is remained as ES5, as `escomplex-js` depends on `esprima` 1 which doesn't support ES6. </del> While `lib` is changed to ES6, as Node 10.9+ already has full ES6 support, see https://node.green/

### DIFF
--- a/.github/workflows/codestyle.yml
+++ b/.github/workflows/codestyle.yml
@@ -13,7 +13,7 @@ jobs:
     - name: Use Node.js
       uses: actions/setup-node@v1
       with:
-        node-version: 12.x
+        node-version: 14.x
 
     - run: npm ci
     - run: npm run code-style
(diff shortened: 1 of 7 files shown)


ariya:
Superseded by #2086.

PR #2062, opened 2020-12-08
Move complexity check to ESLint

### LINKED ISSUE #2059: infrastructure: move test/check-complexity to ESLint complexity check
https://github.com/jquery/esprima/blob/master/test/check-complexity.js calculates cyclomatic complexity for every function, and prints out the most complex 6 functions.

If any function has complexity > 24, the test would fail.

There are several drawbacks here:
* the tool check-complexity.js relies on, https://github.com/philbooth/escomplex-js, is out of maintenance. All the siblings (forks https://github.com/jared-stilwell/escomplex-js, origins https://github.com/escomplex/escomplex) are out of maintenance too. It doesn't support ES6 (as it depends on esprima 1.x), which means we cannot generate ES6 target if we stick on it.
* It checks the generated output js instead of the source.
* the work done is mostly a lint thing.

As ESLint already natively supports limiting complexity, better to move to ESLint, which is already used in this repo.

ljqx:
This PR fixes #2059 

Different from the suggestion in the issue, which uses "complexity" rule built-in in ESLint, this PR creates a new custom rule to match with the original logic we have, which cannot be supported by the "complexity" rule:

* Switch/Case (`options.switchcase = false`) doesn't count.
* Logical operator (`options.logicalor = false`) doesn't count ([&&](https://github.com/philbooth/escomplex-ast-moz/blob/efdd455ec3e5e4f8a66c3266ee3c1f1ae77625b7/src/syntax/LogicalExpression.js), ||, ?? though existing version doesn't support ??).


ariya (approved):
Very good idea :100:

PR #2076, opened 2021-04-12
bugfix import

jogibear9988:


jogibear9988:
@ariya @ljqx could you review?

don't know if this is correct: https://github.com/jquery/esprima/pull/2076/files#diff-abe2818e9c29aadf84401cb7ef8286c7cbce9253eba09f6fa52e22ca4e57b17eR1287

first thought parseUnaryExpression is correct, but import('a' + 'b') is valid. but parseExpression is wrong, cause then import('a', 'b') would not fail

PR #2086, opened 2021-05-19
use modern nodejs versions

Meir017:
for reference - https://nodejs.org/en/about/releases/
image


ariya:
I definitely agree with the addition of v16, but was there a specific reason to drop v8 and v6 checks?
As long as it doesn't cause a hardship on our part, I don't think we should actively exclude testing those versions.

Meir017:
@ariya done

Issue #2028, opened 2020-06-29
Esprima demo interface seems to be broken


goelayu:
### Steps to reproduce
1) Go to https://esprima.org/demo/parse.html#
2) The cursor location seems to be incorrect 

```js
esprima.parse('some code')
```

ariya:
@goelayu Any information on the web browser and operating system used to reproduce this issue?

nanyang24:
@ariya 
I made a gif to preview this issue: 


![Kapture 2020-07-17 at 15 21 12](https://user-images.githubusercontent.com/17287124/87759737-5902b300-c841-11ea-9e7b-9881c34cb844.gif)

ariya:
The editor in the demo is [github.com/Microsoft/monaco-editor](https://github.com/Microsoft/monaco-editor), so likely it's related to that.
Can you try with the online [Monaco demo](https://microsoft.github.io/monaco-editor/playground.html)? Does it behave differently with another browser?

nanyang24:
> The editor in the demo is [github.com/Microsoft/monaco-editor](https://github.com/Microsoft/monaco-editor), so likely it's related to that.
> Can you try with the online [Monaco demo](https://microsoft.github.io/monaco-editor/playground.html)? Does it behave differently with another browser?

Yes, I have tried again to make sure I only see the problem in the Esprima demo.
And I use Monaco Editor a lot in my work, maybe I can help you? 
But I don't actually see the original code for the demo in this repository, could you point me to it?

ariya:
> And I use Monaco Editor a lot in my work, maybe I can help you?
> But I don't actually see the original code for the demo in this repository, could you point me to it?

Check the [gh-branches](https://github.com/jquery/esprima/tree/gh-pages). Maybe we just need an updated Monaco editor, and if you can help with that, it'll be fantastic!

PR #2095, opened 2021-06-11
Fix curser misplacement in editor

nanyang24:
Fixed bug reported in https://github.com/jquery/esprima/issues/2028

After my testing, the cursor misplacement problem only occurs in macOS Chrome. So the solution is that something is interfering with Monaco's cursor positioning.

image

The global CSS base configuration interferes with the Monaco editor fonts.

So, We should specify FontStyle in Monaco options, not in our CSS.

After testing, it works well.

### DIFF
--- a/demo/components.js
+++ b/demo/components.js
@@ -304,6 +304,7 @@ Demo = (function () {
                 language: language,
                 folding: false,
                 lineNumbersMinChars: 3,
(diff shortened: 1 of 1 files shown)


ariya:
Great work @nanyang24, thank you!

PR #2099, opened 2021-07-10
Fix missing optional attributes from tree JSON files

lahma:
Latest merge was missing some optional attributes from JSONs which were introduced by a PR before that, fixed missing attributes.


ariya:
Thank you @lahma 👍

PR #2105, opened 2021-11-03
Fix bigint and seperator

jogibear9988:


jogibear9988:
retrigger

jogibear9988:
i could not use real bigint type or we could not compile to es5

PR #2106, opened 2021-11-03
support for es2022 class propertydefinitions and private members ->

jogibear9988:
see https://github.com/estree/estree/blob/master/es2022.md


jogibear9988:
it also removes asyncFunctionexpression and asyncArrowFunctionExpression, cause they are not in estree, async is a property on the functionExpression.
we also don't have extra objects for generatorFunctions

jogibear9988:
already merged in esprima net

